PR #3, opened 2018-04-24
Add .mjs as file extension for JavaScript

robertrossmann:
`.mjs` is becoming the standard extension for Node.js ES2015 module system.

### DIFF
--- a/assets/database_languages.json
+++ b/assets/database_languages.json
@@ -113,7 +113,8 @@
   {
     "language": "Javascript",
     "extensions": [
-      "js"
+      "js",
+      "mjs"
     ]
   },
   {


boyter:
Looks good. Only catch, the file to change is actually ~/languages.json in the root. I have merged anyway and then added it in there and then run `go generate` to actually build it in see commit 3a1162c59b23ba3419fa2ba6df49b50c5111b928 for this one.

robertrossmann:
Oh, did not see that coming! Thanks for merging so fast! 🚀

boyter:
No problem. I have updated the README to point out how to do this in the future as well.

boyter:
See https://github.com/boyter/scc#addingmodifying-languages for future reference, not that you will need it anymore but in case someone stumbles on this.

PR #8, opened 2018-06-05
validate languages.json during constants generation

nemith:
I was trying to add a new language definition to languages.json but, as
is with json, the pesky commas and other syntax was getting in the way.
When running `go generate` invalid JSON is encoded into constants.go and
then scc would just silently igore the error.  As a result no language
definitions would load.

This add some basic JSON validation to the go generate process as well
as exposes some other possible errors.  It also will generate the
constants to a temporary file before copying over making the process
more atomic and not clobbering a file on a failure.

Since there is now json validation a panic as added to scc itself to
barf if it can't read the json data.

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -39,8 +39,14 @@ var LanguageFeatures = map[string]LanguageFeature{}
 func processConstants() {
 	var database map[string]Language
 	startTime := makeTimestampMilli()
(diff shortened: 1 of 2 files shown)


boyter:
Neat! One of those things that in hindsight I should have added a while ago. Merging in.

PR #10, opened 2018-06-05
Add thrift as a language

nemith:
.thrift files are IDL definitions for thrift (similar to protobufs).
The spec can be found here https://thrift.apache.org/docs/idl


boyter:
Thanks for the submission. Always nice to get additional languages supported.

Issue #17, opened 2018-06-27
API support


jhaynie:
I've been looking into your repo and would like to see if you might be interested in turning this into an API -- such that you could use it as a library in addition to a command line program.

Looking through your code, it seems like we could generalize the FileJob into a more generic type or interface and make `countStats` public and it could easily then be called as a library.

My use case is that I'd like to be able to feed a reader into `countStats` so that I can dynamically pipe a source file (in my case from memory) into your API and have it return me the results.

Is this something you'd be interested discussing further?

boyter:
Sounds reasonable. It shouldn't be too much work to do so.

As you say all you would really need to do is expose `countStats` and job done. Although you would probably want to add in the language check stuff as well as `countStats` has no concept of what language the file is. So long as you stick to using the inbuilt collection it should work though.

Did you want to send through a PR for this? Im happy to do it for you when I get the time, but a PR might be faster.

jhaynie:
I don't mind seeing if we can create a PR for you to review around this. I'm hoping to try and POC it next week

boyter:
No worries. Ill have a look myself. I never considered that someone might want to do this, and I was learning Go at the time so I didn't really write it with that in mind. Be interested to see what someone who actually knows Go does with it :)

boyter:
Looks like this was merged in without issue. If you need anything else let me know, otherwise ill close this down in a few days.

jhaynie:
Sweet! awesome. i'm going to try and open source the little tool we built (as part of a much larger pipeline) here in the next few days and i'll ping you with it.

boyter:
I just did some internet stalking, and only just found out who you are. Very cool that you would be using this tool that I wrote for something!

jhaynie:
haha LOL. thanks.  I just opened source the code as a library which is using scc here: https://github.com/pinpt/ripsrc

would it be possible for you to do a release so I can get this building with the new PR (vs my fork)?

thanks!

boyter:
No problem, https://github.com/boyter/scc/releases/tag/v1.4.0

Ill be on a flight for the next 14 hours or so so hopefully that's all you need. Let me know if there is any issues and we can work through them in 24 hours or so.

boyter:
@jhaynie Cool tool. I noticed you were adding license detection in there as well. Interestingly I did a lot of work there as well, https://github.com/boyter/lc/ although the code there is a much more rough state, perhaps there is something you can use. 

[...]

PR #18, opened 2018-06-28
Proposed changes to support API

jhaynie:
Here's a quick PR (related to #17) that i worked with this afternoon that is minimalistic. I experimented with passing in an `io.Reader` instead of a `[]byte` but it was a bit invasive so I decided to hold off.

As part of this change, I'd like to support an optional callback that can be set on the FileJob that allows you to get per line parsing information which is important for my use case.  If not provided, works the same as today.

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -4,10 +4,10 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"runtime"
-	"strings"
 	"io/ioutil"
+	"runtime"
 	"sort"
+	"strings"
 )
 
 // Flags set via the CLI which control how the output is displayed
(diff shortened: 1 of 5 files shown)


jhaynie:
@boyter ok, i was able to replace our internal tool with scc last night and run some good benchmarks this morning and this increases our processing by 2.5-3x on average.  appreciate if you could review and if you're ok, merge.

boyter:
Looks good to me. I had a try with it to check if performance was impacted by the change and thankfully CPU branch predictor wins again.

If you get the time in the future an example I can add to the docs would be useful, but otherwise I will do it at some point. Id love to point at the repo you are using (if public) in the future as an example of someone using it though.

PR #28, opened 2018-09-10
bitmask prefilter for code scanning

dbaggerman:
Use a bitmask pre-filter before entering string/comment checks, in a
way conceptually similar to bloom filters.

Based on simple testing on my laptop, this reduces `scc` of the linux-4.19-rc1 codebase from ~7.5s to ~4s.

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -88,6 +88,29 @@ func checkForMatchMultiOpen(currentByte byte, index int, endPoint int, matches [
 	return 0, nil
 }
 
+func checkForMatchMultiClose(currentByte byte, index int, endPoint int, matches []OpenClose, fileJob *FileJob) int {
+
+	potentialMatch := true
+	for i := 0; i < len(matches); i++ {
+		if currentByte == matches[i].Close[0] {
+			potentialMatch = true
+
+			for j := 1; j < len(matches[i].Close); j++ {
+				if index+j > endPoint || matches[i].Close[j] != fileJob.Content[index+j] {
+					potentialMatch = false
+					break
+				}
+			}
+
+			if potentialMatch {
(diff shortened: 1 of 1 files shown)


dbaggerman:
Oops, I didn't notice I was working against an older checkout.
However, rebasing against the current master seems to make this slower.
The same benchmark gives me ~6.5s with the current upstream master, and ~5s with a  rebased version of this change.

boyter:
Oh! Neat!

Just what I was looking for. Would you mind if I hold off for a few days for this one? I am in the middle of tweaking how it runs now on a branch and have managed to improve the runtime like so,

```
Benchmark #1: ./scc1.9.0 linux
  Time (mean ± σ):      2.343 s ±  0.097 s    [User: 27.740 s, System: 0.868 s]
  Range (min … max):    2.187 s …  2.509 s
```

to 

```
Benchmark #1: ./scc1.10.0 linux
  Time (mean ± σ):      1.392 s ±  0.019 s    [User: 19.415 s, System: 0.825 s]
  Range (min … max):    1.367 s …  1.430 s
```

So it looks like the above coupled with the changes I am working on and we should be able to get this below 1 second on a 16 core linux box.

dbaggerman:
Redid the rebase, favouring the other branch and cherry picking a couple of optimisations from the current HEAD. It now runs in 3.5s, a respectable improvement over the 4s I had before.

boyter:
Breaks a few of the tests though. Ill have a look into those once I finish with my branch.

boyter:
Have merged in the latest changes I was working on to improve speed. Probably breaks this totally, but then again the tests were broken anyway. Having a look at it now to see if applying it on top of the new changes a boost.

PR #33, opened 2018-09-13
implement tries for token matching

dbaggerman:
Replaces matching of [][]byte slices with a trie / prefix tree.
Invalidates the need for pre-checking with bitmasks or otherwise.

Improves performance by ~15% in my testing:

 4 cores, master: Time (mean ± σ):      6.360 s ±  0.007 s    [User: 24.677 s, System: 0.679 s]
 4 cores, tries:  Time (mean ± σ):      5.489 s ±  0.008 s    [User: 21.145 s, System: 0.690 s]

 8 cores, master: Time (mean ± σ):      3.217 s ±  0.005 s    [User: 24.840 s, System: 0.708 s]
 8 cores, tries:  Time (mean ± σ):      2.784 s ±  0.005 s    [User: 21.270 s, System: 0.733 s]

16 cores, master: Time (mean ± σ):      1.660 s ±  0.016 s    [User: 24.936 s, System: 0.778 s]
16 cores, tries:  Time (mean ± σ):      1.446 s ±  0.014 s    [User: 21.378 s, System: 0.801 s]

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -69,57 +69,33 @@ func ProcessConstants() {
 
 	startTime = makeTimestampMilli()
 	for name, value := range database {
-		complexityChecks := [][]byte{}
-		complexityMask := byte(0)
-		singleLineComment := [][]byte{}
-		singleLineCommentMask := byte(0)
-		multiLineComment := []OpenClose{}
-		multiLineCommentMask := byte(0)
-		stringChecks := []OpenClose{}
-		stringMask := byte(0)
-		processMask := byte(0)
+		complexityTrie := &Trie{}
+		slCommentTrie := &Trie{}
+		mlCommentTrie := &Trie{}
+		stringTrie := &Trie{}
 
 		for _, v := range value.ComplexityChecks {
-			complexityChecks = append(complexityChecks, []byte(v))
-			complexityMask |= v[0]
+			complexityTrie.Insert([]byte(v))
 		}
-		processMask |= complexityMask
 
 		for _, v := range value.LineComment {
-			singleLineComment = append(singleLineComment, []byte(v))
-			singleLineCommentMask |= v[0]
+			slCommentTrie.Insert([]byte(v))
 		}
-		processMask |= singleLineCommentMask
 
 		for _, v := range value.MultiLine {
-			multiLineComment = append(multiLineComment, OpenClose{
-				Open:  []byte(v[0]),
-				Close: []byte(v[1]),
-			})
-			multiLineCommentMask |= v[0][0]
+			mlCommentTrie.InsertClose([]byte(v[0]), []byte(v[1]))
 		}
-		processMask |= multiLineCommentMask
 
 		for _, v := range value.Quotes {
-			stringChecks = append(stringChecks, OpenClose{
-				Open:  []byte(v[0]),
-				Close: []byte(v[1]),
-			})
-			stringMask |= v[0][0]
+			stringTrie.InsertClose([]byte(v[0]), []byte(v[1]))
 		}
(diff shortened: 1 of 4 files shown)


boyter:
Oh now this is a nice one. Will have a large test of this myself.

boyter:
Trying it out on a few repositories now. Looks like this might work better on some situations than others. For example,

```
hyperfine "scc.exe -c C:\Users\bboyter\Documents\Projects\cpython"
Benchmark #1: scc.exe -c C:\Users\bboyter\Documents\Projects\cpython
  Time (mean ± σ):     250.2 ms ±   4.3 ms    [User: 2.6 ms, System: 7.6 ms]
  Range (min … max):   246.8 ms … 261.0 ms
```

vs

```
hyperfine "scc.exe -c C:\Users\bboyter\Documents\Projects\cpython"
Benchmark #1: scc.exe -c C:\Users\bboyter\Documents\Projects\cpython
  Time (mean ± σ):     302.6 ms ±   4.3 ms    [User: 0.0 ms, System: 1.2 ms]
  Range (min … max):   295.2 ms … 311.9 ms
```

Need to do more tests to see which situations this occurs.

boyter:
Similar on Linux

```
$ hyperfine 'scc -c ~/Projects/cpython'
Benchmark #1: scc -c ~/Projects/cpython
  Time (mean ± σ):     481.4 ms ±  18.4 ms    [User: 1.100 s, System: 2.306 s]
  Range (min … max):   467.8 ms … 518.9 ms
```

vs

```
$ hyperfine 'scc -c ~/Projects/cpython'
Benchmark #1: scc -c ~/Projects/cpython
  Time (mean ± σ):     526.2 ms ±  14.5 ms    [User: 1.334 s, System: 2.235 s]
  Range (min … max):   502.6 ms … 560.8 ms
```

My guess is that perhaps with the bitmask back in place it might be faster, as the overhead of the trie might be enough to drag this case down.

boyter:
My artificial benchmark shows that the results are the same between the different versions. Going to investigate what situations this runs faster.

dbaggerman:
I had been testing without the `-c` flag. Given that the complexity checks tend to have more cases, they're likely to benefit more than strings and comments which generally only have one or two cases to match against.

boyter:
Yep that makes sense.

I am still profiling this. It might be possible to get the best of both worlds by putting in the bit-masks for some of the checks. It increases the work for the worst case of a real match slightly but the best case which is nothing matching should be improved.

dbaggerman:
No worries. I have a couple of follow up ideas I want to explore around trie usage, particularly merging the different tries and storing the token type in the leaves so that we perform fewer overall traversals. I'll keep the idea of reinstating the bitmasks in mind.

boyter:
Tried adding bitmask filter back on codeState

```
Benchmark #1: scc.exe -c C:\Users\bboyter\Documents\Projects\cpython
  Time (mean ± σ):     234.3 ms ±   4.3 ms    [User: 3.6 ms, System: 6.1 ms]
  Range (min … max):   228.9 ms … 240.9 ms
```

Linux

```
$ hyperfine 'scc -c ~/Projects/cpython'
Benchmark #1: scc -c ~/Projects/cpython
  Time (mean ± σ):     505.2 ms ±  14.1 ms    [User: 1.034 s, System: 2.401 s]
  Range (min … max):   490.1 ms … 536.4 ms
```

Looks like having both produces a good outcome. Seeing as the application spends most of its time not moving between states because there is nothing to do it makes sense to sell out the move state to speed up the best case.

boyter:
image

It is getting close to the holy grail of not being CPU bound anymore though.

PR #35, opened 2018-09-14
small cpu/mem optimizations

jhaynie:
don't run md5 hash for files that have duplicate turned off
don't create empty byte allocation for completion of the hash (nil suffices as termninator)
set file.content to nil instead of empty byte allocation to free memory in the end


boyter:
Looks good. Nice pickups.

PR #42, opened 2018-10-23
convert from urfave/cli to spf13/cobra

dbaggerman:
Since changes to flags were required to make the change anyway, I also
made a pass to make the flags more consistent with each other and with
similar tools.


boyter:
Looks good!

Issue #44, opened 2018-11-13
Doesn't exit with non-zero exit code on flag error


rbradford:
```
$ scc --bad-flag . 
Error: unknown flag: --bad-flag
Usage:
  scc [flags]

Flags:
      --avg-wage int          average wage value used for basic COCOMO calculation (default 56286)
      --binary                disable binary file detection
      --by-file               display output for every file
      --cocomo                remove COCOMO calculation output
      --debug                 enable debug output
      --exclude-dir strings   directories to exclude (default [.git,.hg,.svn])
      --file-gc-count int     number of files to parse before turning the GC on (default 10000)
  -f, --format string         set output format [tabular, wide, json, csv] (default "tabular")
  -h, --help                  help for scc
  -i, --include-ext strings   limit to file extensions [comma separated list: e.g. go,java,js]
  -l, --languages             print supported languages and extensions
  -c, --no-complexity         skip calculation of code complexity
  -d, --no-duplicates         remove duplicate files from stats and output
  -M, --not-match string      ignore files and directories matching regular expression
  -o, --output string         output filename (default stdout)
  -s, --sort string           column to sort by [files, name, lines, blanks, code, comments, complexity] (default "files")
  -t, --trace                 enable trace output. Not recommended when processing multiple files
  -v, --verbose               verbose output
      --version               version for scc
  -w, --wide                  wider output with additional statistics (implies --complexity)

$ echo $?
0

```

boyter:
Nice pickup. Thanks for submitting the PR for it as well. Have merged in.

PR #45, opened 2018-11-13
Exit with non-zero outcome on command error

rbradford:
If invoking rootCmd.Execute() fails, e.g. due to a flag parsing error. Then
exit with a non-zero exit code.

Fixes: #44

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"os"
+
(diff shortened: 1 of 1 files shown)


boyter:
Very nice. I was wondering if the change to the arguments would cause any issues other than the breaking change. Thanks for fixing it.

PR #46, opened 2018-12-10
Add multi-processor functionality.

kdimak:


boyter:
Rather an interesting looking change. Not sure what it was trying to achieve though?

PR #49, opened 2019-01-09
Lazy loading

boyter:
Modify scc to only load language features as required rather then doing all 222 in a single pass. 
Since the loading of features is a fairly heavy task the savings for the majority of cases is worth it. 
This is especially noticeable when working on smaller repositories.

Version 2.1.0 (with the update) vs 2.0.0 with the full load.

```
$ hyperfine './scc redis-5.0.2' && hyperfine 'scc redis-5.0.2'
Benchmark #1: ./scc redis-5.0.2

  Time (mean ± σ):      81.6 ms ±   5.0 ms    [User: 173.8 ms, System: 265.4 ms]

  Range (min … max):    75.5 ms …  97.1 ms

Benchmark #1: scc redis-5.0.2

  Time (mean ± σ):     124.4 ms ±   2.4 ms    [User: 168.6 ms, System: 289.1 ms]

  Range (min … max):   120.0 ms … 128.4 ms
```

NB this should be backwards compatible with library users as it is only enabled via the main.go file.

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -9,7 +9,6 @@ import (
 	"runtime/debug"
 	"sort"
 	"strings"
-	"sync"
 )
 
 // Flags set via the CLI which control how the output is displayed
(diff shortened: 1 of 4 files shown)


processor/processor.go
```diff
@@ -68,70 +78,113 @@ func ProcessConstants() {
...
-			Tokens:                tokenTrie,
-			Nested:                value.NestedMultiLine,
-			ComplexityCheckMask:   complexityMask,
-			MultiLineCommentMask:  multiLineCommentMask,
-			SingleLineCommentMask: singleLineCommentMask,
-			StringCheckMask:       stringMask,
-			ProcessMask:           processMask,
+	// Check if already loaded and if so return because we don't need to do it again
+	LanguageFeaturesMutex.Lock()
+	_, ok := LanguageFeatures[loadName]
```
dbaggerman: I wonder if there's any argument for using a [syncmap](https://godoc.org/golang.org/x/sync/syncmap) here. Not sure if there'd be enough contention on the lock to justify it, but you never know.

boyter: I did look into that briefly at some point. At the time it was slower than just putting in a mutex. In this case I was being a little lazy though because it was easier to just add it in then change everything around it. 

At some point in the future it might be worth trying it out again as it would make the code simpler which I am always in favor of.

dbaggerman: I tried it quickly and got the same result of being very slightly slower.

Issue #54, opened 2019-02-11
No counts for single file


pombredanne:
See https://raw.githubusercontent.com/corretto/corretto-8/da7e192d1f7a73340cb1ec9b2f08645f8d189769/src/jdk/src/share/native/sun/java2d/cmm/lcms/cmscgats.c for instance

```
$ wget https://raw.githubusercontent.com/corretto/corretto-8/da7e192d1f7a73340cb1ec9b2f08645f8d189769/src/jdk/src/share/native/sun/java2d/cmm/lcms/cmscgats.c
$ ./scc cmscgats.c
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
───────────────────────────────────────────────────────────────────────────────
Total                        0         0        0         0        0          0
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop $0
Estimated Schedule Effort 0.000000 months
Estimated People Required NaN
───────────────────────────────────────────────────────────────────────────────
```
but
```
$ perl ../cloc-1.80.pl cmscgats.c 
       1 text file.
       1 unique file.                              
       0 files ignored.

github.com/AlDanial/cloc v 1.80  T=0.02 s (63.5 files/s, 179074.4 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C                                1            811            212           1796
```

boyter:
Oh, the issue would be because you are specifying the file. If you remove `cmscgats.c` from your argument it should work. EG.

```
$ wget https://raw.githubusercontent.com/corretto/corretto-8/da7e192d1f7a73340cb1ec9b2f08645f8d189769/src/jdk/src/share/native/sun/java2d/cmm/lcms/cmscgats.c
--2019-02-12 09:48:13--  https://raw.githubusercontent.com/corretto/corretto-8/da7e192d1f7a73340cb1ec9b2f08645f8d189769/src/jdk/src/share/native/sun/java2d/cmm/lcms/cmscgats.c
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.28.133
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.28.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 74850 (73K) [text/plain]
Saving to: ‘cmscgats.c’

[...]

PR #55, opened 2019-02-12
Accept a single file as a command line argument #54

pombredanne:
This patch allow to run the scc CLI with a single file path as an argument
rather than a directory.
I explicitly license my contribution under a choice MIT or The Unlicense.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/processor/file.go
+++ b/processor/file.go
@@ -76,6 +76,8 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 
 	var isSoloFile bool = false
 	var all []os.FileInfo
+	// clean path including trailing slashes
+	root = filepath.Clean(root)
 	target, _ := os.Lstat(root)
 	if !target.IsDir() {
 		// create an array with a single FileInfo
--- a/processor/file.go
+++ b/processor/file.go
@@ -5,6 +5,7 @@ import (
 	"github.com/karrick/godirwalk"
 	"github.com/monochromegane/go-gitignore"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 	"runtime/debug"
@@ -72,7 +73,18 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 	totalCount := 0
 
 	var wg sync.WaitGroup
-	all, _ := ioutil.ReadDir(root)
+
+	var is_solo_file bool = false
+	var all []os.FileInfo
+	target, _ := os.Lstat(root)
+	if !target.IsDir() {
+		// create an array with a single FileInfo
+		all = append(all, target)
+		is_solo_file = true
+	} else {
+		all, _ = ioutil.ReadDir(root)
+	}
+
 	// TODO the gitignore should check for further gitignores deeper in the tree
 	gitignore, gitignoreerror := gitignore.NewGitIgnore(filepath.Join(root, ".gitignore"))
 	resetGc := false
@@ -83,6 +95,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 		regex = regexp.MustCompile(Exclude)
 	}
 
+	var fpath string
 	for _, f := range all {
 		// Godirwalk despite being faster than the default walk is still too slow to feed the
 		// CPU's and so we need to walk in parallel to keep up as much as possible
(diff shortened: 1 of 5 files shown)


processor/file.go
```diff
@@ -72,7 +73,18 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 	totalCount := 0
 
 	var wg sync.WaitGroup
-	all, _ := ioutil.ReadDir(root)
+
+	var is_solo_file bool = false
```
dbaggerman: Snake case isn't very common in Go, `isSoloFile` would be more consistent with the rest of the codebase.

However, there is actually a simpler way to solve this. Simply replacing

    all, _ := ioutil.ReadDir(root)

with

    rootFileInfo, _ := os.Lstat(root)
    if !rootFileInfo.IsDir() {
            root = filepath.Dir(root)
    }
    all := []os.FileInfo{rootFileInfo}

is the only change necessary in `processor/file.go`. That lets us better reuse the existing logic below for handling files vs directories.

Other than that, it looks good to me. Thanks for the PR.

pombredanne: > Snake case isn't very common in Go, isSoloFile would be more consistent with the rest of the codebase.

My snakyness shows up right of the bat... Thank you for the much appreciate review

pombredanne: re
 ```
rootFileInfo, _ := os.Lstat(root)
if !rootFileInfo.IsDir() {
        root = filepath.Dir(root)
}
all := []os.FileInfo{rootFileInfo}
```
> is the only change necessary in processor/file.go. That lets us better reuse the existing logic below for handling files vs directories.

That's much cleaner alright!

pombredanne: @dbaggerman Your alternate solution no longer works with directories.

dbaggerman: Oh, good catch. I tested `./scc .` which did work, but other directories don't.

I realise now that the `if !rootFileInfo.IsDir()` condition shouldn't be there. Try

    rootFileInfo, _ := os.Lstat(root)
    // filepath.Clean will remove trailing folder slashes, which otherwise change the behaviour of filepath.Dir
    root = filepath.Dir(filepath.Clean(root))
    all := []os.FileInfo{rootFileInfo}

That works for all the cases I can think of. See if it you can find a way to break it.

dbaggerman: Actually, one issue with this is that ignoring the error on `os.Lstat` now causes a panic if an invalid file is given, instead of the old behaviour of just giving empty results. We'll have to catch that now and exit the goroutine gracefully.

pombredanne: > Actually, one issue with this is that ignoring the error on os.Lstat now causes a panic if an invalid file is given, instead of the old behaviour of just giving empty results. We'll have to catch that now and exit the goroutine gracefully.

@dbaggerman I would have no idea where to start to do this... But there is another issue. You approach works... BUT IMHO no longer walks in parallel as the input for a dir is now a dir and not its list of child directories.

pombredanne: Snaykyness removed and I am waiting for your feedback to my last point to make any further change

dbaggerman: Yeah, you're right about it not walking in parallel with my approach. It already isn't always properly parallel (for example when all the code is in one subdirectory), but I don't want to make the general case worse. Might be worth looking at redesigning that, but I won't hold up your PR.

I'll have one more look when I've got time, but I did notice an obscure error case that your new validation isn't catching:

    ./scc main.go/                     
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x54936c]

When a file exists, it passes your test but later gets processed as a directory which is invalid.

pombredanne: > ./scc main.go/                     
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x54936c]

@dbaggerman fixed in latest commit. Tell me if you want some more refactoring still!

dbaggerman:
Ok, looks good to me now. Thanks @pombredanne.

pombredanne:
@dbaggerman thanks ++ :bowing_man:

Issue #57, opened 2019-02-24
Unable to tell the difference between Coq and Verilog 


boyter:
This is becoming more of an issue, especially with the news that V Lang is also going to use `v` along with Coq and Verilog https://github.com/vlang-io/V/issues/6

Annoying because its quite a large refactor to accommodate this. More to the point is how to implement it? One thought is to scan the first 500 bytes or so and look for tokens that indicate one language or another. This is likely to be problematic however. 

Might have to have a look at how https://github.com/vmchale/polyglot does it and copy whatever it is doing.

Things to keep in mind. Need to do this for every file that has the extension up to some point. As such it needs to be fast. One possible thing to do here is after checking some amount of files if all are determined to be Coq perhaps flip over to just assuming that will be the case for everything else from that point on.

boyter:
A quick check of polyglot https://github.com/vmchale/polyglot/blob/b688f66543327603341f3c2047ff26c57e557ba7/DATS/shared.dats#L356 indicates that keywords is exactly how it determines between languages with the same extension. 

https://github.com/github/linguist apparently also supports this case and should be checked to see how it does it.

boyter:
A quick check of polyglot https://github.com/vmchale/polyglot/blob/b688f66543327603341f3c2047ff26c57e557ba7/DATS/shared.dats#L356 indicates that keywords is exactly how it determines between languages with the same extension. 

https://github.com/github/linguist apparently also supports this case and should be checked to see how it does it.

boyter:
Started working on this https://github.com/boyter/scc/tree/issue57

boyter:
PR for this https://github.com/boyter/scc/pull/58

PR #58, opened 2019-02-26
Issue 57

boyter:
Solves #57 

Allows different languages to share the same extension by looking for case sensitive keywords. This will slow down look-ups in these cases but it should not be such a huge issue in practice. The output showed it taking < 1 millisecond per file, and something like 60,000 nanoseconds when I tested. Something to improve in the future though.

Should in theory be backwards compatible with anyone calling into the process using CountStats as well so no issues there. 

Also added in V language https://vlang.io/ support while doing this (based on Go). 

Once merged going to make this into a new release since its a pretty nice piece of functionality and something that https://github.com/vmchale/polyglot has which this should.


boyter:
Oh Go 1.12 is out as well. Will be able to see if we get some speed boost out of it.

boyter:
```
$ hyperfine './scc ~/redis-5.0.2/' && hyperfine 'scc ~/redis-5.0.2/'
Benchmark #1: ./scc ~/redis-5.0.2/
  Time (mean ± σ):      68.9 ms ±   1.9 ms    [User: 132.3 ms, System: 268.0 ms]
  Range (min … max):    66.4 ms …  75.6 ms

Benchmark #1: scc ~/redis-5.0.2/
  Time (mean ± σ):      72.2 ms ±   6.5 ms    [User: 117.5 ms, System: 282.7 ms]
  Range (min … max):    66.7 ms …  97.2 ms
```
A little more speed out of 1.12 which is nice.

boyter:
Coverage decreased (-0.007%) to 91.366% really coveralls? -0.007?

processor/structs.go
```diff
@@ -48,6 +50,7 @@ type FileJobCallback interface {
 // FileJob is a struct used to hold all of the results of processing internally before sent to the formatter
 type FileJob struct {
 	Language           string
+	Languages          []string // Used to hold potentially more than one language which populates language when determined
```
dbaggerman: I'm not a fan of having `Language` and `Languages` in the same struct. I think we should differentiate the names if they're both required - `PossibleLanguages` and `ActualLanguage` or something.

processor/file.go
```diff
@@ -183,8 +185,11 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 						totalCount++
 						mutex.Unlock()
 
-						LoadLanguageFeature(language)
-						output <- &FileJob{Location: fpath, Filename: f.Name(), Extension: extension, Language: language}
+						for _, l := range language {
+							LoadLanguageFeature(l)
+						}
+
+						output <- &FileJob{Location: fpath, Filename: f.Name(), Extension: extension, Language: language[0], Languages: language}
```
dbaggerman: If this code is no longer responsible for determining the "real" language, I'd argue for not setting the `Language` field here (or on line 265). I'd keep the responsibility in one place, probably in `workers.go` with the condition to guess languages.

processor/workers.go
```diff
@@ -260,6 +262,11 @@ func CountStats(fileJob *FileJob) {
 		return
 	}
 
+	// If we have multiple languages then we need to guess which one this might be
+	if len(fileJob.Languages) != 1 {
```
dbaggerman: I know it _shouldn't_ happen, but I'd still be more defensive here of `len == 0`. Slices can also be `nil`, but that might be a safer / more reasonable assumption.

processor/structs.go
boyter: `Language` will need to remain. This is because it keeps backwards compatibility with those calling direct into CountStats, of which a few applications do now.

Perhaps `PossibleLanguages` over `Languages` then. I'll make that change now.

processor/file.go
boyter: That's a fair point. I initially kept it the same to avoid further processing down the line as it knows the language at this point.

Ill have a look if there is any performance impact from this (though I suspect there should be none).

processor/workers.go
boyter: Sorry not following. You mean check if it is empty? IE

```
if len(fileJob.Languages) == 0 {
```

dbaggerman:
Just stylistic comments so far. I'll take a closer look when I've got time.

One thought though, what are the odds of the same extension meaning different things in the same project? If we guess `.v` to mean verilog in one file, could we cache that and assume that other `.v` files are also verilog? Would that even be a worthwhile performance boost?

processor/workers.go
dbaggerman: Yeah, that's what I meant.

boyter:
That's cool. I have refactored slightly.

I did mention that possibility that on the linked issue actually, https://github.com/boyter/scc/issues/57 

Personally I doubt there will be any mixed case. I suspect adding some threshold IE if 20 files are all of one type then lets assume all will be that type might be a good solution. I also think perhaps just checking the first 1000 bytes of the file might be good enough, but I need to try with more real world tests. Sadly all the languages this is built to deal with are not mainstream which makes this harder.

I think it might be acceptable to push this first with the goal to speed it up afterwards (this is up for debate) as I would want `scc` to support V when it released next month.

processor/workers.go
boyter: Refactored it out totally. Changed method name to `determineLanguage` and gave it a quick out. This was done to accommodate removal of setting `Language` based on `PossibleLanguages`

boyter:
A benchmark against the coq repository with ~1600 coq files shows no noticeable slowdown. This should represent on of the worst cases as it should be hitting the language determination code fairly heavily.

```
$ hyperfine 'scc -c' && hyperfine 'tokei' && hyperfine 'loc' && hyperfine '~/polyglot'
Benchmark #1: scc -c
  Time (mean ± σ):     225.2 ms ±   5.2 ms    [User: 330.0 ms, System: 1203.5 ms]
  Range (min … max):   212.3 ms … 231.8 ms

Benchmark #1: tokei
  Time (mean ± σ):     304.7 ms ±   2.6 ms    [User: 442.0 ms, System: 1147.2 ms]
  Range (min … max):   301.5 ms … 309.6 ms

Benchmark #1: loc
  Time (mean ± σ):     255.1 ms ±  28.6 ms    [User: 529.2 ms, System: 1111.0 ms]
  Range (min … max):   213.6 ms … 300.2 ms

Benchmark #1: ~/polyglot
  Time (mean ± σ):      1.063 s ±  0.037 s    [User: 988.6 ms, System: 1394.1 ms]
  Range (min … max):    1.009 s …  1.136 s
```

Still faster than everything else in this case. I will have a look at the profile for this but at least from a far higher level the price paid for this is not very high.

boyter:
Quick profile, it almost has no real impact. I might tweak it to use indexed loops over range just to see if that cuts some allocations.

image

boyter:
Running against a repository without the need to trigger the logic at all. No impact which is nice.

image

boyter:
No difference being reported between range vs for loop with index. I suspect that perhaps the new GC changes in Go 1.12 might be helping with this. As expected flame graph shows all the time is spent in strings.Index

boyter:
Slight tweak. Reduced the number of characters checked. No longer will it scan the whole file, but only the first 2000 characters. On this repository https://github.com/coq/coq it reduced the % cost by about 50% over checking everything else. I did try 1000 characters which reduced it even further without any accuracy cost but I am not sure if thats a good amount to use as someone may have a heavily commented file header.

boyter:
Debating if adding a cache is worth it. The bookkeeping involved might be higher overhead then is worth it especially with it selling out the worst case of a mixed repository which just happens to contain Coq Verilog and V in it. I will need to find a larger repository than the one I have been testing to see if it is worth the cost. 

I would argue at this point given that `scc` is still outperforming all the other tools while being accurate it is not.

boyter:
Debating if adding a cache is worth it. The bookkeeping involved might be higher overhead then is worth it especially with it selling out the worst case of a mixed repository which just happens to contain Coq Verilog and V in it. I will need to find a larger repository than the one I have been testing to see if it is worth the cost. 

I would argue at this point given that `scc` is still outperforming all the other tools while being accurate it is not.

processor/processor.go
```diff
@@ -122,7 +122,13 @@ func ProcessConstants() {
 	startTime := makeTimestampNano()
 	for name, value := range languageDatabase {
 		for _, ext := range value.Extensions {
-			ExtensionToLanguage[ext] = name
+			_, ok := ExtensionToLanguage[ext]
+
+			if ok {
```
dbaggerman: This condition looks superfluous. `append`ing to a `nil` slice will create a new slice so the

    ExtensionToLanguage[ext] = append(ExtensionToLanguage[ext], name)

line should be all we need.

dbaggerman:
That's the last concern from me. Looks good!

processor/processor.go
boyter: Ah neat. Never thought about that but it makes perfect sense.

boyter:
Cool. Ill make that final change and merge in.

PR #60, opened 2019-02-28
Simplify getExtension with filepath.Ext

apocelipes:
The original method of splitting a string with a "." to get a file extension in `getExtension` is complicated and unintuitive, so I reimplemented this function with filepath.Ext.


boyter:
Oh very nice. I had a feeling there was something wrong when I wrote that. Thanks for the PR.

PR #66, opened 2019-04-02
.gitignore improvements

boyter:
The current .gitignore logic was pretty crappy, to the point it was effectively useless. I noticed that `scc` was needlessly walking things that should have been ignored and then usually discarding the results due to the whitelist.

This PR makes it work at the root level with the idea of building on it so that we can implement recursive .gitignore checks throughout the tree, and perhaps include hg ignore files in the future too. In short it is not meant to be the be all end all solution but at least gets the basics right.

In some cases this will improve performance and in others it will be worse, but more importantly it will improve relevance.

```
$ hyperfine './scc' 'scc'
Benchmark #1: ./scc
  Time (mean ± σ):      33.2 ms ±   2.2 ms    [User: 32.6 ms, System: 102.4 ms]
  Range (min … max):    30.3 ms …  41.9 ms

Benchmark #2: scc
  Time (mean ± σ):      38.1 ms ±   3.4 ms    [User: 31.6 ms, System: 104.6 ms]
  Range (min … max):    34.3 ms …  53.8 ms

Summary
  './scc' ran
    1.15x faster than 'scc'

$ hyperfine 'scc' '~/scc'
Benchmark #1: scc
  Time (mean ± σ):     114.7 ms ±   1.9 ms    [User: 225.0 ms, System: 531.2 ms]
  Range (min … max):   111.5 ms … 119.7 ms

Benchmark #2: ~/scc
  Time (mean ± σ):      56.1 ms ±   2.3 ms    [User: 92.4 ms, System: 195.8 ms]
  Range (min … max):    53.2 ms …  67.7 ms

Summary
  '~/scc' ran
    2.04x faster than 'scc'
```

Importantly while making this all sorts of tests broke, which was rather reassuring to me. I fixed all the logic errors eventually and now `test-all.sh` runs as expected.

```
$ ./test-all.sh
Running go fmt...
Running unit tests...
?       github.com/boyter/scc   [no test files]
ok      github.com/boyter/scc/processor 1.812s
?       github.com/boyter/scc/scripts   [no test files]
Building application...
Running integration tests...
Error: unknown flag: --not-a-real-option
Usage:
  scc [flags]

Flags:
      --avg-wage int          average wage value used for basic COCOMO calculation (default 56286)
      --binary                disable binary file detection
      --by-file               display output for every file
      --cocomo                remove COCOMO calculation output
      --debug                 enable debug output
      --exclude-dir strings   directories to exclude (default [.git,.hg,.svn])
      --file-gc-count int     number of files to parse before turning the GC on (default 10000)
  -f, --format string         set output format [tabular, wide, json, csv] (default "tabular")
  -h, --help                  help for scc
  -i, --include-ext strings   limit to file extensions [comma separated list: e.g. go,java,js]
  -l, --languages             print supported languages and extensions
  -c, --no-complexity         skip calculation of code complexity
  -d, --no-duplicates         remove duplicate files from stats and output
  -M, --not-match string      ignore files and directories matching regular expression
  -o, --output string         output filename (default stdout)
  -s, --sort string           column to sort by [files, name, lines, blanks, code, comments, complexity] (default "files")
  -t, --trace                 enable trace output. Not recommended when processing multiple files
  -v, --verbose               verbose output
      --version               version for scc
  -w, --wide                  wider output with additional statistics (implies --complexity)

PASSED invalid option test
PASSED no directory specified test
PASSED directory specified test
PASSED multiple options test
PASSED regular expression ignore test
PASSED shared extension test 1
PASSED shared extension test 2
PASSED shared extension test 3
PASSED concurrency issue test
PASSED file specified test
PASSED duplicates test
Cleaning up...
=================================================
ALL TESTS PASSED
=================================================
```

### DIFF
--- a/processor/file.go
+++ b/processor/file.go
@@ -91,6 +91,13 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 
 	// TODO the gitIgnore should check for further gitignores deeper in the tree
 	gitIgnore, gitIgnoreError := gitignore.NewGitIgnore(filepath.Join(root, ".gitignore"))
+	if Verbose {
+		if gitIgnoreError == nil {
+			printWarn(fmt.Sprintf("found and loaded gitignore file: %s", filepath.Join(root, ".gitignore")))
+		} else {
+			printWarn(fmt.Sprintf("no gitignore found: %s", filepath.Join(root, ".gitignore")))
+		}
+	}
 
 	resetGc := false
 
@@ -126,7 +133,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 				}
 			}
 
-			if gitIgnoreError != nil || gitIgnore.Match(filepath.Join(root, f.Name()), true) {
+			if gitIgnoreError == nil && gitIgnore.Match(filepath.Join(root, f.Name()), true) {
 				if Verbose {
 					printWarn("skipping directory due to git ignore: " + filepath.Join(root, f.Name()))
 				}
@@ -163,7 +170,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 
 			shouldSkip := false
 
-			if gitIgnoreError != nil || gitIgnore.Match(fpath, false) {
+			if gitIgnoreError == nil && gitIgnore.Match(fpath, false) {
 				if Verbose {
 					printWarn("skipping file due to git ignore: " + f.Name())
 				}
--- a/processor/file.go
+++ b/processor/file.go
@@ -76,7 +76,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 
 	var wg sync.WaitGroup
 
-	var isSoloFile = false
+	isSoloFile := false
 	var all []os.FileInfo
 	// clean path including trailing slashes
 	root = filepath.Clean(root)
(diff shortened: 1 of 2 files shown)


boyter:
NB the drop in coverage is due to adding some more verbose output details. All the important logic is still tested. Probably the main reason I find coverage statistics rather pointless most of the time. Anyway it should be safe to ignore and ill ensure its raised when I get around to adding the additional .gitignore logic.

processor/file.go
```diff
@@ -76,7 +76,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 
 	var wg sync.WaitGroup
 
-	var isSoloFile bool = false
+	var isSoloFile = false
```
dbaggerman: If we're cleaning this up, either just `var isSoloFile bool` (`false` is the zero value for bools) or `isSoloFile := false` would be my preference.

boyter: Ah good point. I just did the refactor thing in the IDE and it picked that up. Not sure why it didn't go the whole way though.

boyter: Done.

dbaggerman:
Other than that small style change, looks good to me.

Issue #61, opened 2019-03-01
Allow passing multiple directories


jackwilsdon:
For example;

```Shell
scc src/ include/
```

Currently `scc` just uses the first file;

https://github.com/boyter/scc/blob/2a32681c7376895fc8454487f6e2ccb50b2b9d3e/main.go#L20-L25

https://github.com/boyter/scc/blob/2a32681c7376895fc8454487f6e2ccb50b2b9d3e/processor/processor.go#L90-L91

https://github.com/boyter/scc/blob/2a32681c7376895fc8454487f6e2ccb50b2b9d3e/processor/processor.go#L314

jackwilsdon:
I've slapped together an unpolished and mostly untested version of this which _seems_ to work fine;

```Diff
diff --git a/processor/file.go b/processor/file.go
index 70a3419..527513f 100644
--- a/processor/file.go
+++ b/processor/file.go
@@ -201,7 +201,6 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 	}
 
 	wg.Wait()
-	close(output)
 	if Debug {
 		printDebug(fmt.Sprintf("milliseconds to walk directory: %d", makeTimestampMilli()-startTime))
 	}
diff --git a/processor/processor.go b/processor/processor.go
index 1fed4fc..0e72a25 100644
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -311,11 +311,18 @@ func Process() {
 	if len(DirFilePaths) == 0 {
 		DirFilePaths = append(DirFilePaths, ".")
 	}
-	fpath := filepath.Clean(DirFilePaths[0])
 
-	if _, err := os.Stat(fpath); os.IsNotExist(err) {
-		fmt.Println("file or directory does not exists: " + fpath)
-		return
+	fpaths := make([]string, 0, len(DirFilePaths))
+
+	for _, fpath := range DirFilePaths {
+		cfpath := filepath.Clean(fpath)
+
+		if _, err := os.Stat(cfpath); os.IsNotExist(err) {
+			fmt.Println("file or directory does not exists: " + cfpath)
+			return
+		}
+
+		fpaths = append(fpaths, cfpath)
 	}
 
 	SortBy = strings.ToLower(SortBy)
@@ -330,7 +337,22 @@ func Process() {
 	fileReadContentJobQueue := make(chan *FileJob, FileReadContentJobQueueSize) // Files ready to be processed
[...]

PR #69, opened 2019-04-17
Add support to run over multiple files or directories

boyter:
Implements request https://github.com/boyter/scc/issues/61 where you can pass in multiple files or directories on the command line and scc will process them as instructed. Biggest change is that the file walker is no longer responsible for closing its stream. This makes the code that spawns them a little more ugly since it now needs to have a waitgroup to track this but I don't seen an obvious way to improve it.

Appears to work over everything I tried. There may be the old ulimit issue again if the directories passed in are large enough, but I was not able to trigger it on my machine.

### DIFF
--- a/processor/file.go
+++ b/processor/file.go
@@ -218,7 +218,7 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 	}
 
 	wg.Wait()
-	close(output)
+
 	if Debug {
 		printDebug(fmt.Sprintf("milliseconds to walk directory: %d", makeTimestampMilli()-startTime))
 	}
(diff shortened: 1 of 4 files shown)


boyter:
Ill be away till around the 29th of May so if you are happy merge, if not ill look after any feedback when I get back. No issues either way.

dbaggerman:
Looks good to me. Having a monitor goroutine close a `chan` when a waitgroup completes is a pattern I've used for dealing with multiple writers. I usually prefer to break out explicit manager/worker functions rather than using nested inline `go func()` but it's not a big deal.

Issue #67, opened 2019-04-16
Allow passing multiple regex (--not-match)


andyfitzgerald:
Consider accepting multiple regular expressions instead of just one. Example

``` sh
scc --not-match ".*\.csv" --not-match "specialDirectory/.*\.txt" .
```

Currently `scc` just uses the last specified regex (--not-match) option.

boyter:
Sounds interesting. Would it not be acceptable in this case to have them combined into a single regex though?

```
scc --not-match ".*\.csv|specialDirectory/.*\.txt" .
```

The above is not tested BTW

andyfitzgerald:
That's certainly an option; however, I propose this as an enhancement to improve readability and maintainability of the code invoking `scc`. 

Imagine a project that embeds `scc` in a script (make, bash, etc). There will likely be a few variables, or an array variable, containing special cases to ignore. Please see a contrived example below.

``` make
SCC_REGEX_EXCLUDES := \
    ".*\.csv" \
    "specialDirectory/.*\.txt" \
    ".*\build.*" \
    "otherSpecialDirectoryPattern" \
    ".*\.hex" \
    "more and more tedious regex.*"
SCC_EXCLUDES_REGEX_STR := $(addprefix --not-match=, $(SCC_REGEX_EXCLUDES))

...

target:
    scc $(SCC_EXCLUDES_REGEX_STR) directoryToScan
```

In this way, it's easier to read and maintain the script. It also simple to add cases for developers who are not regex experts.

You could argue that the encapsulating script could assemble a regex in a way that's better for maintainability, but I would counter that it's prone to error.

boyter:
That's fair enough. I tend to do multiples when using grep in some situations as well. I am not that familiar with cobra which does the argument parsing but I suspect this should be possible.

andyfitzgerald:
Assuming that you mean the command line arguments to `scc`, my guess would be that you could re-use the solution for #61 or what is currently done for `--exclude-dir` arguments. Multiple `--exclude-dir` arguments appears to work in my local tests.

boyter:
Neat. I'm having a look at this now.

boyter:
So this works very differently to the way #61 does because that worked outside of Cobra. Not sure if https://github.com/spf13/cobra supports multiple of the same arguments to it. I didn't see anything on the main page anyway. Will have to look into that further.

dbaggerman:
[...]

PR #70, opened 2019-04-30
Make --exclude a StringArray argument

dbaggerman:
This is to address #67. It changes the `--not-match` argument to a `StringArray` to allow multiple exclusions to be passed.

I used `StringArray` rather than `StringSlice` since the latter splits on commas, and while I disapprove of commas in file names I think it's better to be tolerant. Besides, comma separating regular expressions could get ugly.

### DIFF
--- a/main.go
+++ b/main.go
@@ -104,11 +104,11 @@ func main() {
 		false,
 		"remove duplicate files from stats and output",
 	)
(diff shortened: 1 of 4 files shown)


boyter:
Looks good to me. Ill add some tests into the bash test script when I get a chance to play with it.

Issue #72, opened 2019-05-02
First comment line in C# is counted as Code


ptupitsyn:
**Describe the bug**
When first line in C# file contains single-line comment (starts with `//`), it is counted as code.

**To Reproduce**

1. Download file 
[FooClass.zip](https://github.com/boyter/scc/files/3137556/FooClass.zip)
2. Extract
3. Scc

**Expected behavior**
```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
C#                           1        14       11         2        1          0
───────────────────────────────────────────────────────────────────────────────
Total                        1        14       11         2        1          0
───────────────────────────────────────────────────────────────────────────────
```

**Actual behavior**
```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
C#                           1        14       12         1        1          0
───────────────────────────────────────────────────────────────────────────────
Total                        1        14       12         1        1          0
───────────────────────────────────────────────────────────────────────────────
```

**Desktop**
 - Windows 10

boyter:
This is an unusual one. When run via command line it does indeed fail. With the following test however,

```
func TestCountStatsIssue72(t *testing.T) {
	ProcessConstants()
	fileJob := FileJob{
		Language: "C#",
	}

	fileJob.Content = []byte(`// Comment 1
namespace Baz
{
    using System;

    public class FooClass
    {
        public void Test(string report)
        {
          // Comment 2
          throw new NotImplementedException();
        }
    }
}`)

	CountStats(&fileJob)

	if fileJob.Lines != 14 {
		t.Errorf("Expected 14 lines")
	}

	if fileJob.Code != 11 {
		t.Errorf("Expected 11 lines got %d", fileJob.Code)
	}

	if fileJob.Comment != 2 {
		t.Errorf("Expected 2 lines got %d", fileJob.Comment)
	}

	if fileJob.Blank != 1 {
		t.Errorf("Expected 1 lines got %d", fileJob.Blank)
	}
}
```

[...]

PR #74, opened 2019-05-02
Add test case and temp fix for issue72

boyter:
Don't actually want to merge this yet, as I belive the solution should be more generic but this does work to fix the issue in https://github.com/boyter/scc/issues/72

I am thinking that perhaps the full table of BOM https://en.wikipedia.org/wiki/Byte_order_mark#Byte_order_marks_by_encoding should be included in this to ensure that it works for all encoding types. No idea what impact that would have though. Is it even worth doing it for all of them?

Looking for ideas on this before it gets merged in for good to ensure its a solution that does not need to be revisited if possible.

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -35,8 +35,8 @@ const (
 )
 
 // Taken from https://en.wikipedia.org/wiki/Byte_order_mark#Byte_order_marks_by_encoding
+// These indicate that we cannot count the file correctly so we can at least warn the user
 var ByteOrderMarks = [][]byte{
-	{239, 187, 191},       // UTF-8 make this first as it is likely to be the most common case
 	{254, 255},            // UTF-16 BE
 	{255, 254},            // UTF-16 LE
 	{0, 0, 254, 255},      // UTF-32 BE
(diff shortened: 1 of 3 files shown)


dbaggerman:
Checking for the BOM seems reasonable to me. Maybe look at https://golang.org/pkg/bytes/#HasPrefix for checking it though, rather than checking individual bytes.

Non UTF-8 encoding probably aren't ASCII-ish enough for `scc` to read them properly anyway. Maybe we could just use their presence to show a warning/explanation to the user about encoding support.

dbaggerman:
Although, I suppose if we wanted to we could use `https://golang.org/pkg/unicode/utf16` and `https://golang.org/pkg/encoding/binary` to populate a `Trie` with utf-16 encoded tokens and get UTF-16 support that way.

Not sure it's worthwhile, but it's interesting nonetheless.

boyter:
I do like that it works on bytes at the moment, which was a deliberate choice to avoid ever worrying about the encoding (just didn't expect BOM for the most part) which is just a painful thing to work with without something like Python's unicode dammit.

I have the table mostly done now,

```
var ByteOrderMarks = [][]byte{
	{239, 187, 191},       // UTF-8 make this first as it is likely to be the most common case
	{254, 255},            // UTF-16 BE
	{255, 254},            // UTF-16 LE
	{0, 0, 254, 255},      // UTF-32 BE
	{255, 254, 0, 0},      // UTF-32 LE
	{43, 47, 118, 56},     // UTF-7
	{43, 47, 118, 57},     // UTF-7
	{43, 47, 118, 43},     // UTF-7
	{43, 47, 118, 47},     // UTF-7
	{43, 47, 118, 56, 45}, // UTF-7
	{247, 100, 76},        // UTF-1
	{221, 115, 102, 115},  // UTF-EBCDIC
	{14, 254, 255},        // SCSU
}
```

As you suggested have flipped over to https://golang.org/pkg/bytes/#HasPrefix I had a feeling something was there but wanted to see if I could resolve it first before looking at the API. Cheers for finding it.

boyter:
Seems to happen more than you would expect 

image

boyter:
I think this is much closer to a proper solution. Going to add more tests for it to cover all the cases first.

boyter:
I think this is much closer to a proper solution. Going to add more tests for it to cover all the cases first. Will then check for any performance issues, which I doubt will be a problem.

boyter:
`github.com/boyter/scc/processor.checkBomSkip (0.18%, 0.02s)` 

when tested over a very large repository

```
Language                 Files     Lines     Code  Comments   Blanks Complexity
Total                    40167   9506667  5699791   2839927   966949     751032
```

So next to no impact. Took me several runs to get the profiler to even pick it up based on the default sampling size actually.

boyter:
Pretty happy with this now. 

@dbaggerman feel free to review at any point.

dbaggerman:
Since we know that most of these encodings won't actually scan correctly, we should show a warning/error when we detect them in the BOM, not just skip the BOM and carry on.

A token like `for {` will encode to the same sequence of bytes in ASCII or utf8, so skipping the BOM and carrying on is correct. But for most/all other encodings it will encode to a different byte sequence so the complexity analysis would never be correct. Even single byte/character tokens like quotes or newlines might not be the same in other encodings so even the line count could be incorrect.

dbaggerman:
Actually, it just occurred to me that utf16 would almost certainly contain bytes containing zero and fail the `isBinary` check.

boyter:
Yeah the isBinary does break on those. I agree, should change this to emit a warning if the BOM is found for anything non UTF-8 and then proceed as normal. Ill refactor.

boyter:
Modified as per above. Only skips if UTF-8 BOM is found. Otherwise if `-v` verbose option is enabled it will print a warning indicating the file might not be counted correctly. 

Nice side effect is that this should reduce that run-time overhead as the second check only needs to happen if verbose is enabled in which case printing will probably be slower than the check itself.

dbaggerman:
Ok, looks good to me.

Issue #73, opened 2019-05-02
Blanks are not counted correctly in Java


ptupitsyn:
**Describe the bug**
In some cases blank lines are not counted correctly in Java files.
I've simplified the example somewhat, but could not pinpoint problematic lines.

**To Reproduce**

1. Download file [AbstractSACParser.zip](https://github.com/boyter/scc/files/3137601/AbstractSACParser.zip)
2. Extract
3. Scc

**Expected behavior**
```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
Java                         1       139      103        16        20          8
```

**Actual behavior**
```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
Java                         1       139      119        15        5          8
```

**Desktop**
 - Windows 10

boyter:
Does not appear to be related to #72 as the file is UTF-8 encoded.

boyter:
Issue is the following line,

```
if (c == '\\' || (c == '\"')) {
```

The " is being picked up as the start of a string which is throwing out the count. Because `scc` counts blank lines as being code if inside a string unlike tokei this is why the count is incorrect.

In this case we could check for \ before the string since this should only occur inside a char. However not all languages work like that. I think this is acceptable for the moment as a quick fix though and to create a new issue to evaluate how this works in other languages. It should work for most of the common languages though which is a good short term win.

boyter:
Fix for this case exists on branch https://github.com/boyter/scc/pull/75

boyter:
Merged into master.

This will still need work to make it more flexible, so no release till that is done in https://github.com/boyter/scc/pull/76

That said closing down because if you build from master its all fine now and will make it into the next release.

PR #75, opened 2019-05-05
Resolve issue73 where quotes in character would trigger string state

boyter:
A fix for https://github.com/boyter/scc/issues/73

If a " was used inside a character it would trigger the string state. Quick fix is to check for \ before it and stay in the code state. 

Long term fix is to have a state for dealing with quotes, perhaps inside strings as they have the same rules. However this requires updating the whole JSON document and checking that quotes work like strings in all languages (unlikely). This seems like a good reasonable short term fix for most C like languages.

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -142,7 +142,12 @@ func codeState(
 
 			switch tokenType, offsetJump, endString := langFeatures.Tokens.Match(fileJob.Content[i:]); tokenType {
 			case TString:
-				currentState = SString
+				// It is safe to -1 here as to enter the code state we need to have
+				// transitioned from blank to here hence i should always be >= 1
+				// This check is to ensure we aren't in a character declaration
+				if fileJob.Content[i-1] != '\\' {
+					currentState = SString
+				}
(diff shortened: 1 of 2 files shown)


dbaggerman:
I'm a little concerned that not all languages follow exactly the same quoting rules, and hardcoding the backslash like this might break other languages.

PowerShell for example uses a backtick rather than a backslash to escape characters, as far as I'm aware.

There are also cases like backtick strings in Go, where escape characters can't be used at all.

    fmt.Println(`\`)

will print `\`

boyter:
Absolutely. Its a short term thing that should catch most cases for C like languages but can then be improved with https://github.com/boyter/scc/pull/76 This is to get it in place then improve. I also want to change the return from the states to a struct now because its getting hard to read, but thats also going on that other branch.

In your Go example it would hit the string state first and be counted correctly so no issues there thankfully.

Certainly not intending to get this into a release, however it lays the ground work for the other branch.

dbaggerman:
Agreed that it probably isn't ready for a release as is, but as a starting point it's fine.

PR #81, opened 2019-06-17
Clean up

boyter:
Not quite ready for review, mostly here so I can see what im doing more easily, oh and so I get travis to do some builds so I can see if I broke things.

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -133,18 +133,18 @@ func stringState(sta *state) (int, int64) {
 	return sta.index, sta.currentState
 }
 
-func codeState(sta *state, digest *hash.Hash) (int, int64, []byte, [][]byte, bool) {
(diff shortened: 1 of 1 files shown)


dbaggerman:
Thanks for the push notification spam!

Joking aside, I'm not a fan of the `sta` variable name. I think I'd go with making the struct name more explicit (e.g. `scannerState`), and calling the variable `state` (or `scanner`?).

If we're trimming down to a single argument everywhere, there's also the option of using method syntax. `state.stringState()` might be a bit opaque, but something like `scanner.scanString()` doesn't seem too bad.

boyter:
Hahaha yeah I forgot it would email you a lot. Sorry about that.

Yep I don't like the name either. Once I finish this off ill change it.

Issue #82, opened 2019-06-19
false statistics about a language that is not actually present in repo.


fahadhk:
**Describe the bug**
When I run `scc` a command line directly from repo directory i am getting correct statistics but when I run it like mentioned below I also get ASP.NET in statistics. How come that can happen?

**To Reproduce**

1. using `scc` as command line.
2. command line arguments. `$ scc /path/to/a/repo`

**Expected behavior**
I expect it to show statistics about Golang, markdown, makefile and gitignore only.

**Desktop (please complete the following information):**
 - OS: [Linux Ubuntu 19.04]

dbaggerman:
I'm able to reproduce this. Looks like `.git` directories aren't being filtered out when using an absolute path.

ASP.NET lists `.master` as one of its file extensions, and we're also incorrectly picking up files in `.git` called `master` without a file extension. Removing the `.git` subdirectory from a repo and running with an absolute path doesn't show any ASP.NET.

fahadhk:
Thanks for feedback. 

Any fix coming up for this?

dbaggerman:
This is the first time it's been reported, so I'm still looking into it. I don't think it'll be too hard to fix though, we should have it patched soon.

boyter:
Thanks for reporting this. Could have sworn I verified this sometime back. I suspect I broke it when working on the .gitignore fixes a few months back.

```
# Regression issue https://github.com/boyter/scc/issues/82
a=$(./scc . | grep Total)
b=$(./scc ${PWD} | grep Total)
if [ "$a" == "$b" ]; then
    echo -e "${GREEN}PASSED git filter"
else
    echo -e "${RED}======================================================="
    echo -e "FAILED git filter"
    echo -e "================================================="
    exit
fi
```

A test replicating the issues which I dropped into `test-all.sh`. @dbaggerman let me know if you don't have the time to deal with this and ill look at it.

boyter:
@fahadhk thanks to the excellent work of @dbaggerman this looks like it has been resolved. If you build from master you should be able to confirm it. If so feel free to close this, otherwise ill do it in a few days.

boyter:
Closing down as promised.

fahadhk:
@boyter Thanks for closing and for the fix. I was not so active in the last few days.

boyter:
No problem.

PR #83, opened 2019-06-20
fix ignoring .git on absolute paths

dbaggerman:


dbaggerman:
This should fix #82

boyter:
Looks good to me.

Issue #79, opened 2019-05-07
Verilog files identified as V


boyter:
**Describe the bug**

Verilog files are being identified as V.

**To Reproduce**

Clone https://github.com/seldridge/verilog.git
Run scc over the folder `./src` inside it
Observe the output where V language is identified

```
$ scc
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
───────────────────────────────────────────────────────────────────────────────
SystemVerilog                7       598      383       171       44         66
V                            6       965      771       143       51         36
───────────────────────────────────────────────────────────────────────────────
Total                       13      1563     1154       314       95        102
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop $31,398
Estimated Schedule Effort 4.118292 months
Estimated People Required 0.903127
───────────────────────────────────────────────────────────────────────────────
```

**Expected behavior**

All the files in this directory should be identified as Verilog.

**Desktop (please complete the following information):**
 - OS: Window, Linux, MacOS
 - Version 2.4.0

rtennill:
The files were picked up as Verilog in my pre-merge testing. Should be an easy thing to verify. 
  eab82433a904010e41804768a94ac2ee1e7c7e76 and 0d1f113accbad40721925c94737ca5942538adef

boyter:
Changes to fix the Coq verification as linked appears to not have regressed this.

PR #89, opened 2019-07-24
Hdl improvements

rtennill:
address #79 for admittedly limited test cases and also reset terminal colors at test exit.

The extension .sv is recommended for System Verilog as some FPGA tools will not recognize .v as System Verilog.


boyter:
Looks good to me. Thanks for the submission. Merging in.

Issue #88, opened 2019-07-24
Feature request: add yaml format option that mirrors cloc --yaml


rtennill:
This option would make scc a drop-in replacement for cloc in automated flows.

boyter:
Sounds reasonable to me.

boyter:
I have merged in your PR. Ill do some minor work on it as mentioned there.

Could you possibly write a small case of where you would use this to be a drop in for cloc? I would love to add it to the README.md so everyone else gets the benefit as well :)

boyter:
So looking at this a bit further, would be it better to make this option something such as `cloc-yaml`? If it is set to YAML it precludes having a custom YAML format later which would include all of the files themselves similar to the JSON output?

I think this makes a lot of sense and makes it more obvious that this is a compatibility for cloc. Thoughts @rtennill ?

rtennill:
Sure, that sounds totally reasonable. I have some WIP on another branch that touches the yaml output since it looks like I dropped a commit somehow. I'll adjust to follow the notion of cloc-yaml.

boyter:
Sounds good. Im going to refactor master to use that then.

boyter:
Have refactored master to use `cloc-yaml` but also made it support `cloc-yml` just so it works as most would expect without surprise.

boyter:
@rtennill Any word? Id love to get this pushed out as a release as the new Go compiler provides another 3% performance boost which is always nice :)

rtennill:
borked my vm somehow...will get back to this ASAP. Sorry

boyter:
No problem. I am having a look at the output now and getting it to match my version of cloc.

Do you have some sort of validator to confirm the format?

rtennill:
Nothing so fancy. I was just eyeballing it against cloc output from the included examples and vhdl/verilog files from a Xilinx example project.

That brings up an interesting question...my cloc is quite old and I haven't checked to see if the format changed....probably ought to do a quick sanity check there.

boyter:
No problem.

I have already updated it a bit to ensure its closer to the actual format. Ill see what I can do about adding a parser as well.

boyter:
@rtennill Could you try building from master and let me know how it works for you as a drop in replacement now?

```
$ scc --format cloc-yaml processor
# https://github.com/boyter/scc/
header:
  url: https://github.com/boyter/scc/
  version: 2.6.0
  elapsed_seconds: 0.007
  nFiles: 17
  nLines: 5219
[...]

PR #90, opened 2019-07-24
Cloc format

rtennill:
First cut at addressing #88. 

Probably not the best way to do it but this is first experience with go. Suggestions welcomed!

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,7 +86,7 @@ func main() {
 		"format",
 		"f",
 		"tabular",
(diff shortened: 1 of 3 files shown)


boyter:
Looks pretty good. I think ill have to work on getting that coverage number up myself since its just a personal thing that I want it over 90%.

I wonder if anyone is going to complain that this expects `yaml` and not `yml`.

I am going to merge because all works when I made a local copy. Ill work on improving the coverage when when its above 90% again I will push out a release.

PR #93, opened 2019-07-29
Recursive ignore file check

boyter:
Adds recursive .ignore/.gitignore file checks to files/directories after the first one.

Should speed up processing in some cases, and slow them in some but more importantly improve accuracy.

This is somewhat in preparation of getting multiple .gitignore file support working, once I figure out how to actually implement it (looking for thoughts here)


processor/file.go
```diff
@@ -162,24 +139,19 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
...
-				if Verbose {
-					printWarn("skipping directory due to git ignore: " + filepath.Join(root, f.Name()))
-				}
-				shouldSkip = true
-			}
-
-			if ignoreError == nil && ignore.Match(filepath.Join(root, f.Name()), true) {
-				if Verbose {
-					printWarn("skipping directory due to ignore: " + filepath.Join(root, f.Name()))
+			for _, i := range ignores {
```
dbaggerman: Given how commonly `i` is used for numeric indexes, I would avoid using it here. `ignore` is clearer and not too long.

processor/file.go
```diff
@@ -162,24 +139,19 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
...
+				if i.Match(filepath.Join(root, f.Name()), true) {
+					if Verbose {
+						printWarn("skipping directory due to ignore: " + filepath.Join(root, f.Name()))
+					}
+					shouldSkip = true
```
dbaggerman: This could use a `break` like the loop above.

processor/file.go
```diff
@@ -206,18 +178,13 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
...
-				if Verbose {
-					printWarn("skipping file due to git ignore: " + f.Name())
-				}
-				shouldSkip = true
-			}
-
-			if ignoreError == nil && ignore.Match(fpath, false) {
-				if Verbose {
-					printWarn("skipping file due to ignore: " + f.Name())
+			for _, i := range ignores {
```
dbaggerman: Given how these loops seem to come as a pair with `for _, exclude := range excludes`, could we unify the two?

It should be possible to define a `FileMatcher` interface, and get both (wrapped) `regexp` matchers and `gitignore` matchers in the same list.

processor/file.go
```diff
@@ -268,7 +235,24 @@ func walkDirectoryParallel(root string, output chan *FileJob) {
 	}
 }
 
-func walkDirectory(toWalk string, blackList []string, extensionLookup map[string][]string) []FileJob {
+func loadIgnoreFile(root string, filename string, ignores []gitignore.IgnoreMatcher) []gitignore.IgnoreMatcher {
+	ig, err := gitignore.NewGitIgnore(filepath.Join(root, filename))
```
dbaggerman: I would use `ignore` here too, over `ig`.

processor/file.go
boyter: That might be one to do when I get around to adding the multiple .gitignore support.

I want to keep them separate for the moment at least till that's done because its implementation is already doing my head in :)

boyter: BTW if you want to help rebuild godirwalk let me know. Will want to make it as fast as possible, run in parallel without having to built in the consumer libraries and allow some sort of pre-flight so we can request it to let us know if .gitignore files exist before processing so it can be loaded for futher processing.

dbaggerman: Sure, I'm happy to help with that. I've thought on and off about handling non-linear streams of work which don't map well to `chan`s (e.g. walking directories and finding more sub-directories in the process). I've got some ideas I want to play around with, I'll probably put something simple/generic up on my github when I get something to work.

boyter: Neat. If its fast (IE almost as fast) as whats there it would be a good thing to switch over to. Those recursive .gitignore files are a bitch with godirwalk.

PR #94, opened 2019-08-03
Refactor file processing

dbaggerman:
Not ready for merge, just putting this up for discussion.

I played around with an idea for a system that lets parallel workers both read from and push work back to a pool of work. I've put a first draft of that up at [dbaggerman/cuba](https://github.com/dbaggerman/cuba).

I took a look at what it would look like to convert `scc` from its current directory scanning to something based on cuba. I haven't really tested it properly, aside from verifying it compiles and gives the same[1] output as the master version of `scc`.

Current performance:

```
Benchmark #1: ./scc-master linux/
  Time (mean ± σ):      1.123 s ±  0.015 s    [User: 13.957 s, System: 0.743 s]
  Range (min … max):    1.102 s …  1.150 s    10 runs
 
Benchmark #2: ./scc-cuba linux/
  Time (mean ± σ):     959.8 ms ±   9.5 ms    [User: 12.859 s, System: 1.116 s]
  Range (min … max):   946.6 ms … 974.6 ms    10 runs
 
Summary
  './scc-cuba linux/' ran
    1.17 ± 0.02 times faster than './scc-master linux/'
```

An isolated apples to apples comparison shows it being slower than `godirwalk` for very simple directory scans, which suggests there may be room for more improvement. The fact that it shares work well across threads probably explains why it did better than `master` above. Since `scc` does more work in the scanner, it will benefit more from the added parallelism.

```
Benchmark #1: cubawalk
  Time (mean ± σ):     176.4 ms ±   6.0 ms    [User: 399.4 ms, System: 210.6 ms]
  Range (min … max):   164.3 ms … 185.3 ms    16 runs
 
Benchmark #2: godirwalk
  Time (mean ± σ):     157.5 ms ±  15.3 ms    [User: 92.7 ms, System: 67.0 ms]
  Range (min … max):   134.6 ms … 181.9 ms    21 runs
 
Benchmark #3: go-filepath-walk
  Time (mean ± σ):     355.5 ms ±  43.3 ms    [User: 186.1 ms, System: 181.6 ms]
  Range (min … max):   290.3 ms … 419.2 ms    10 runs
 
Summary
  'godirwalk' ran
    1.12 ± 0.12 times faster than 'cubawalk'
    2.26 ± 0.35 times faster than 'go-filepath-walk'
```

[1] Scanning the linux repo, it reported one less python file. After some digging, it's excluded by `tools/testing/selftests/tc-testing/.gitignore`.


dbaggerman:
Huh... I deleted a test and still pushed the test coverage up almost a full percent. I guess it's a bit misleading since it's relying on `cuba` code I haven't written any tests for yet, but still interesting.

processor/file_test.go
```diff
@@ -85,8 +84,10 @@ func TestWalkDirectoryParallel(t *testing.T) {
 	GcFileCount = 10
 
 	inputChan := make(chan *FileJob, 10000)
-	walkDirectoryParallel("../", inputChan)
-	close(inputChan)
+
+	dirwalker := NewDirectoryWalker(inputChan)
+	dirwalker.Walk("../")
+	dirwalker.Run()
```
boyter: Maybe return the walker on the walk? Could then chain it.

```
dirwalker := NewDirectoryWalker(inputChan).Walk("../").Run()
```

dbaggerman: That wouldn't work because `Walk` actually returns an error that was being ignored. I've updated the tests to check the returned error value.

dbaggerman:
I've refactored the API slightly and haven't found any glaring issues with `cuba`, so I'll leave it up to you whether you're happy to merge.

boyter:
Going to run it against a few thousand repositories first to ensure every output matches, Assuming all good then ill merge her in.

dbaggerman:
The output may not match exactly when there are nested `.gitignore`/`.ignore` files.

What I saw when I compared the linux repo was this build reporting one fewer python file. There is a `.gitignore` file at `tools/testing/selftests/tc-testing/.gitignore`, which contains a rule for `plugins`. There is one `__init__.py` file in `tools/testing/selftests/tc-testing/plugins` which I believe to be the discrepancy.

But to me, that seems more like the 2.6.0 release not handling nested ignore files than a problem with this PR.

boyter:
The nested .gitignores I can filter out manually. Just want to make sure everything else works on a large set of repos first.

boyter:
Merged in because it seemed all good, but geting a few issues EG

```
$ scc main.go
ERROR 2019-08-14T18:25:28Z: failed to read main.go: readdirent: not a directory───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines     Code  Comments   Blanks Complexity
Go                           1       171      161         4        6          2
Total                        1       171      161         4        6          2
Estimated Cost to Develop $3,969
Estimated Schedule Effort 1.876811 months
Estimated People Required 0.250551
```

Also when running `test-all.sh`

```
=================================================
FAILED git ignore filter
=================================================
```

Which is doing the following,

```
touch ignored.xml
a=$(./scc | grep Total)
b=$(./scc --no-gitignore | grep Total)
```

ignored.xml is in the .gitignore file but the ignore option makes no difference.

boyter:
https://github.com/boyter/scc/issues/99

PR #109, opened 2019-08-29
Update languages.json

anup1920sri:


boyter:
100% agree with @dbaggerman that this cannot be merged as is.

Perhaps you could explain your intent @anup1920sri with this can we can assist with the change?

boyter:
Going to timebox this. If no response in the next 24 hours I will close it.

boyter:
No response closing.

PR #112, opened 2019-09-14
add homebrew to install docs; rearrange install instructions

elindsey:
I added homebrew to the install docs. They were starting to get long, so I rearranged it a bit - let me know if you don't like the restructuring and I'll just tack it on to the existing text instead.

Might be easier to read the formatted version here https://github.com/elindsey/scc/tree/bmaster


boyter:
Thanks for the submission. Approved and merged.

PR #117, opened 2019-10-05
Fix bug 114 and include 115 #! support

boyter:
As it says. This resolves the bug 114 where files named as the language EG `java` would be counted as that language, they now need the appropriate extension.

While I was in there I added #! support. This highlighted a bug with the blacklist which needed to be resolved as all the files in the git directory were being loaded and checked for #! which was incorrect.


processor/detector.go
```diff
@@ -0,0 +1,26 @@
...
+	ProcessConstants()
+
+	if !strings.HasPrefix(content, "#!") {
+		return "", errors.New("Missing #!")
+	}
+
+	for k, v := range ShebangLookup {
+		for _, x := range v {
+			// detects both full path and env usage
+			if strings.Contains(content, "/"+x) || strings.Contains(content, " "+x) {
```
dbaggerman: This looks like it has the potential for false positives. The first case should be checking for the basename specifically, but this could match on directory names (e.g. `/python/perl/ruby` would match any of `python`, `perl` or `ruby` when only `ruby` should be matched.

The second case is only really valid if the first part is the path to an `env` program.

processor/file.go
```diff
@@ -186,6 +187,12 @@ DIRENTS:
 
 func newFileJob(path, name string) *FileJob {
 	extension := ""
+
+	// If there is no . in the filename or it starts with one then check if #! or other
+	if !strings.Contains(name, ".") || (name[0] == '.' && strings.Count(name, ".") == 1) {
```
dbaggerman: It might be marginally more efficient here to call `strings.Count` once and check for zero or one, rather than calling both `strings.Contains` and `strings.Count`.

processor/file.go
```diff
@@ -141,7 +141,8 @@ DIRENTS:
 		isDir := dirent.IsDir()
 
 		for _, black := range PathBlacklist {
-			if strings.HasPrefix(path, filepath.Join(job.root, black)) {
+			//if strings.HasPrefix(path, filepath.Join(path, black)) {
```
dbaggerman: This commented code can be removed.

processor/processor.go
```diff
@@ -309,7 +323,7 @@ func printLanguages() {
 	})
 
 	for _, name := range names {
-		fmt.Println(fmt.Sprintf("%s (%s)", name, strings.Join(database[name].Extensions, ",")))
+		fmt.Println(fmt.Sprintf("%s (%s)", name, strings.Join(append(database[name].Extensions, database[name].FileNames...), ",")))
```
dbaggerman: This use of `append` makes me cringe a little bit. I know `printLanguages` is outside the main loop and the arrays aren't too big, but this is still allocating new backing arrays and copying the elements for each invocation.

processor/workers.go
```diff
@@ -692,6 +691,33 @@ func fileProcessorWorker(input chan *FileJob, output chan *FileJob) {
...
 
 				fileStartTime := makeTimestampNano()
+
+				// Needs to always run to ensure the language is set
+				determineLanguage(res)
+
+				// If the type is #! we should check to see if we can identify
+				if res.Language == SheBang {
+					cutoff := 200
+					if len(res.Content) > cutoff {
```
dbaggerman: Is this back to front? Shouldn't it be checking for content length *less than* the cutoff length?

processor/processor.go
boyter: Yeah I am with you, however keep in mind FileNames is empty most of the time as well.

I guess it could change to have a if condition where it needs to do this, but seems like additional work for no real benefit as you mention this almost never runs. 

Happy to modify it if you can think of something cleaner.

processor/workers.go
boyter: Yep 100% correct. Fixed.

processor/detector.go
boyter: That's a good point. Although... there is a special place in hell for those who organise their paths like that.

The second part is indeed for `env` detection.

Ill have a ponder on the first. I think the check might need to look for the first newline, then check if it has a postfix that matches. The `env` check would remain the same but within those same bounds.

boyter: Although the problem is the fuzzyness of how people can use #!

Consider the following cases, which were all covered using the technique I did initially,

```
"#!/usr/bin/perl"
"#!  /usr/bin/perl"
"#!/usr/bin/perl -w"
"#!/usr/bin/env perl"
"#!  /usr/bin/env   perl"
"#!/usr/bin/env perl -w"
"#!  /usr/bin/env   perl   -w"
"#!/opt/local/bin/perl"
"#!/usr/bin/perl5"
```

I don't even know if the 3rd example `"#!/usr/bin/perl -w"` is legitimate actually. Anyway it did work with all of the above.

Perhaps for the first check removing all white-space and `-w` might be enough to resolve this.

boyter: Updated with slightly modified code that meets all requirements and also deals with newlines. New test cases added.

processor/detector.go
```diff
@@ -0,0 +1,33 @@
...
+	cln = strings.Replace(cln, "-w", "", -1)
+
+	for k, v := range ShebangLookup {
+		for _, x := range v {
+			// detects both full path and env usage
+			if strings.HasSuffix(cln, "/"+x) || strings.Contains(content, " "+x) {
+				return k, nil
+			}
+		}
+	}
```
dbaggerman: Replacing `-w` isn't really appropriate as it's a perl specific argument, not a general shebang thing. The argument after the executable name is completely arbitrary and gets interpreted by the executable.

I think it's worth extracting the executable/arguments out properly, rather than guessing based on substrings.

This should pull both out with a single pass through the string:

```go
func scanShebang(shebang string) (string, string, bool) {
    if !strings.HasPrefix(shebang, "#!") {
        return "", "", false
    }

    slashIndex := 0
    wsStartIndex := 0
    wsEndIndex := 0

    body := strings.TrimSpace(shebang[2:])
    for i := 0; i <  len(body); i++ {
        if body[i] == '/' {
            slashIndex = i
        }

        if body[i] == ' ' || body[i] == '\t' {
            if wsStartIndex == 0 {
                wsStartIndex = i
            }

            wsEndIndex = i
        }

        if wsEndIndex != 0 && wsEndIndex != i {
            break
        }

        if body[i] == '\n' {
            break
        }
    }

    exe := body[slashIndex+1:wsStartIndex]

    arg := ""
    if wsEndIndex != 0 {
        arg = body[wsEndIndex+1:]
    }

    return exe, arg, true
}
```

processor/processor.go
dbaggerman: Yeah, I know it's not in a performance critical code path and the arrays are tiny -- but it still makes me cringe.

Anyway, I can't think of a simple way of doing it without requiring extra code, so I'm not going to argue over it.

boyter: Cool going to resolve this then.

processor/detector.go
boyter: The above fails on a few of my test cases sadly, due to string slicing with wsStartIndex being at 0. I was modifying it to work then realised.

There may be an easier approach. There are only 2 cases that actually matter.

1. if `/env ` is present then the next non-blank string to either a space, newline or tab is the program
2. if `/env ` is not present then the program is whatever follows the last occurrence of `/` to a space, newline or tab.

There are no other cases. The above accounts for the 3 possible cases,

1. `#!/usr/bin/perl -w`
2. `#!/usr/bin/perl`
3. `#!/usr/bin/env perl -w`

It does not matter how many spaces follow the env or if any follow the command.

Ill have a think on this. I have a feeling there is a simple solution here that works for all cases.

dbaggerman: You're right, I didn't properly handle the case of not having an argument when I slapped this together. A condition before/around the `exe := body[slashIndex+1:wsStartIndex]` should be all it needs to fix it though.

> There may be an easier approach. There are only 2 cases that actually matter.

The problem is that you need to match on whitespace, which means that the various `strings.*` functions aren't going to cut it. You could do it with a regular expression, but most people would find `^#![^ \t]*/([^ \t]+)([ \t]+(.*))?$` illegible. Hell, I'm generally pretty good with regular expressions and I think I might struggle with that one if I hadn't just written it.

> 3. `#!/usr/bin/env perl -w`

This case is actually invalid, since shebang parsing doesn't split arguments. Looks like GNU `env` does accept `#!/usr/bin/env --split-string=perl -w`, but I've never seen that form in the wild.

boyter: No I don't mean using regex or string. Another small state machine should do it. Can do the #! check in there as well.

You start in nothing, move to shebang, then to first / where you record the last found one till you hit a space or newline, and as such know exactly where to make the cuts. Very similar to what you did in fact.

boyter: Committed something that is currently broken to demonstrate how it could work. Works in this case for one case at least.

dbaggerman: Oh, I see what you mean. I thought that trying to embed the `env` logic into the state machine would overcomplicate it. But it's a valid solution if you want to go that way.

In your demo sample, `state == 0` and `state == 1` are the same thing. I'm pretty sure the only separate states you need are "before whitespace" and "after whitespace".

boyter: Yep, just a quick thing I threw together. 100% not ready for release till I actually fix it :)

boyter: Have a look at whats in there now @dbaggerman moved to use a state machine. 

Benchmarks look promising and it works with all existing cases

```
$ go test -run=SheBang -bench=.                                                                                                                                                                                     goos: linux                                                                                                                                                                                                         goarch: amd64                                                                                                                                                                                                       pkg: github.com/boyter/scc/processor                                                                                                                                                                                BenchmarkScanSheBangFuzz-8                                                500756              2329 ns/op
```

boyter:
image

Tried this version running on 10 copies of the linux kernel against the usual suspects. Much faster which is nice to see. So even with all the new functionality over the last few months speed has not been impacted in any great way.

processor/detector.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func scanForSheBang(content []byte) (string, error) {
+	state := 0
+	lastSlash := 0
+
+	candidate1 := ""
+	candidate2 := ""
+
+	for i := 0; i < len(content); i++ {
```
dbaggerman: You can write this as `for i := range content {`.

processor/detector.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+
+			if isWhitespace(content[i]) {
+				candidate2 = string(content[lastSlash:i])
+				state = 4
+			}
+		case state == 4:
+			break
+		}
+	}
```
dbaggerman: Putting conditions in the `case` statements is an usual use-case. Go lets you do it, but it's only really meant for more mixed use-cases. The normal syntax (with the variable in-line with the `switch`) looks like:
```
switch state {
case 0:
    // ...
case 1:
    // ...
}
```
The normal syntax allows compiling down to a jump table (1 comparison per loop), while putting conditions in the `case` is basically syntactic sugar for chaining `if` conditions (N comparisons per loop).

processor/detector.go
```diff
@@ -0,0 +1,96 @@
...
+
+	switch {
+	case candidate1 == "env":
+		return candidate2, nil
+	case candidate1 != "":
+		return candidate1, nil
+	}
```
dbaggerman: Same as above. Can be written as `switch candidate1` / `case "env":`.

processor/detector.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return "", errors.New("Count not find")
+}
```
dbaggerman: Looks like you've more or less reimplemented my snippet in a slightly different way. But if this version is easier for you to understand, then that's fine. Other than the code style comments it looks fine to me.

boyter: Yep its pretty much the same thing. I just got it working with all cases is all.

processor/detector.go
boyter: Does it actually go down to a jump table? Last time I checked Ken was on record as saying it won't hence you see some code use a Map with functions. Not that it matters it should be in that format.

processor/detector.go
boyter: Actually I don't think it can due to the `!=` in there. I just thought it was less ugly then two if conditions.

dbaggerman: It could be done as `case "env"`, `case ""`, `default`. But it's not a big deal anyway.

boyter: Yeah its just at the end. Ill leave it as it is for the moment. Can always be refactored later.

Issue #92, opened 2019-07-25
Add minified/generated file check


boyter:
Similar to binary, if the file appears to be minified or generated perhaps don't include it.

boyter:
@dbaggerman Would love some thoughts on the below.

There are a few things to consider here. The first being should be be included by default IE minified files are excluded. I can see this being correct because like .git folders you probably don't want to include them but also wrong because it was committed, and as such is part of the codebase. I am slightly leaning towards it being a flag to opt into.

The second thing is should the files be excluded, or perhaps counted in a different bucket (or optionally both). For example if I have JavaScript and turn the functionality on the result may look like 

```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
JavaScript                  29     13485     2640      2541     8304       1733
(min) JavaScript         29     13485     2640      2541     8304       1733
───────────────────────────────────────────────────────────────────────────────
```

I have prefixed the (min) because for large languages it would otherwise be cut off. Although this makes sorting a bit of a pain... or win as all the minified types would be lumped together.

Being able to exclude the minified files would also be an option but I like the above due to it still counting, just as a different name.

This would work for JavaScript, JSON, XML, CSS and a bunch of other file types.

boyter:
Ideas on how to implement this.

https://twitter.com/schristopheraz/status/1182247287835234304?s=20

```
Line length of line 1 > 50
If file fileName[1]= "min"
If spaces.count() < 10 (lol)
If carriage return count < 3
```

Might work. However I think 50 is too low, 300 might be a better number to avoid false positives.

jhaynie:
we did a lot of testing based on our data set and came up with these numbers. not perfect per se. we'll probably revisit at some point to build a more accurate model but this seems to work well for us: https://github.com/pinpt/ripsrc/blob/master/ripsrc/fileinfo/fileinfo.go#L44

boyter:
I did consider a hard cut byte size as well. Might expand the ticket out to cover "generated" files which that would do. Cheers.

boyter:
Trying out on https://github.com/boyter/scc/tree/minified

Seems a very simple check works quite well for all my test cases so far.

boyter:
Picks up the `constants.go` file in this repository which is pretty neat.

PR #121, opened 2019-10-14
Minified/Generated code check

boyter:
For https://github.com/boyter/scc/issues/92

Allows you to turn on minified/generated code checks which flags the files differently OR exclude them entirely.

The power is yours so choose wisely!


processor/workers.go
```diff
@@ -486,6 +488,15 @@ func CountStats(fileJob *FileJob) {
 		if fileJob.Content[index] == '\n' || index >= endPoint {
 			fileJob.Lines++
 
+			if MinifiedGenerated {
+				if avgLineByteCount == 0 {
+					avgLineByteCount = index - lineOffset
+				} else {
+					avgLineByteCount = (avgLineByteCount + (index - lineOffset)) / 2
```
dbaggerman: This isn't a real average - it's more of a rolling average weighted towards lines at the end of the file.

This doesn't even seem like it's needed. Just stick `avgLineByteCount := len(fileJob.Content) / fileJob.Lines` after the state machine.

processor/workers.go
```diff
@@ -737,6 +767,15 @@ func fileProcessorWorker(input chan *FileJob, output chan *FileJob) {
 					duplicates.mux.Unlock()
 				}
 
+				if IgnoreMinifiedGenerate {
+					if res.Minified {
```
dbaggerman: Purely style preference, but I'd make this `IgnoreMinifiedGenerate && res.Minified` over the nested condition.

processor/workers.go
boyter: Ah the old not notice whats going on when you are doing it. Good point.

processor/workers.go
boyter: 100% agree with you. Will refactor.

processor/workers.go
boyter: The weighted towards the end was useful since some JS files follow that pattern but this might still work.

boyter: Tests pass so all good.

boyter:
If you feel its all good @dbaggerman just merge it. Otherwise ill clean up whatever you throw back asap.

dbaggerman:
Looks good to me.

Issue #123, opened 2019-10-15
Panic when file ends with a block comment and no newline


belak:
**Describe the bug**
When a file ends in a special quoted string and no newline, it will panic. A sample stack trace and test case has been included below.

```
% scc .
panic: runtime error: index out of range [649] with length 649

goroutine 98 [running]:
github.com/boyter/scc/processor.verifyIgnoreEscape(0xc000264000, 0xc000265200, 0xc000264900, 0xc000265b00, 0xc000266400, 0x7f2700237f00, 0x0, 0x0, 0x0, 0xc00019f5e0, ...)
	/private/tmp/scc-20191015-33901-1vdnpeq/scc-2.9.0/src/github.com/boyter/scc/processor/workers.go:352 +0xca
github.com/boyter/scc/processor.blankState(0xc0004a4000, 0x286, 0x288, 0x1, 0xc00051f4a0, 0x0, 0x0, 0xc0000ac2b8, 0x3, 0x8, ...)
	/private/tmp/scc-20191015-33901-1vdnpeq/scc-2.9.0/src/github.com/boyter/scc/processor/workers.go:317 +0x10f
github.com/boyter/scc/processor.CountStats(0xc0004a4000)
	/private/tmp/scc-20191015-33901-1vdnpeq/scc-2.9.0/src/github.com/boyter/scc/processor/workers.go:459 +0xe38
github.com/boyter/scc/processor.fileProcessorWorker.func1(0xc0000a6120, 0xc000236000, 0xc0000a6180, 0xc000236010)
	/private/tmp/scc-20191015-33901-1vdnpeq/scc-2.9.0/src/github.com/boyter/scc/processor/workers.go:743 +0x243
created by github.com/boyter/scc/processor.fileProcessorWorker
	/private/tmp/scc-20191015-33901-1vdnpeq/scc-2.9.0/src/github.com/boyter/scc/processor/workers.go:709 +0xb0
```

**To Reproduce**

In theory this should match the line numbers for if the file ends in a newline, but that doesn't seem to be counted the same and I really didn't want to figure that out. I've included a regression test.

```
func TestCountStatsIssue(t *testing.T) {
	ProcessConstants()
	fileJob := FileJob{
		Language: "Python",
	}

	fileJob.Content = []byte(`"""
hello there! how's it going?
"""`)

	CountStats(&fileJob)

	if fileJob.Lines != 2 {
		t.Errorf("Expected 2 lines got %d", fileJob.Lines)
	}

[...]

PR #124, opened 2019-10-15
Fix panic when file ends with a special quote and no newline

belak:
Fixes issue #123


boyter:
Quality PR. Thanks for the submission @belak

PR #127, opened 2019-10-17
Add ignore large file feature

boyter:


boyter:
Put the suggested stat call into the file portion. Even if it does slow down the file traversal a lot (edge case anyway because this is an option) its worth the memory savings and the burnt CPU time. Of course the line count still needs to happen afterwards but whatever.

The defaults are based on ripsrc.

boyter:
Bloody coverage tool. I wish I could configure it to ignore slight drops in coverage if the total is still over 90% or some such.

dbaggerman:
> Even if it does slow down the file traversal a lot

`DirectoryWalker.Readdir` actually already has access to the file size, which it could push it into `newFileJob` without requiring an additional `os.Stat` call. `dirent` is of type `os.FileInfo` which has a `Size()` getter.

processor/workers.go
```diff
@@ -693,6 +693,13 @@ func fileProcessorWorker(input chan *FileJob, output chan *FileJob) {
 					continue
 				}
 
+				if NoLarge && res.Lines >= LargeLineCount {
+					if Verbose {
+						printWarn(fmt.Sprintf("skipping large file due to line length: %s", res.Location))
+					}
+					continue
+				}
+
```
dbaggerman: Maybe it's worth moving this into `CountStats` and `return` as soon as the line limit is reached? It would save a least some processing rather than waiting until after the file is fully processed.

boyter: I did consider that... I would hope usually the byte limit kicks in first. The outside part is still needed, just it exits earlier. So you get something like

Line 485 in workers.go
```
                // This means the end of processing the line so calculate the stats according to what state
		// we are currently in
		if fileJob.Content[index] == '\n' || index >= endPoint {
			fileJob.Lines++

                        if NoLarge && fileJob.Lines >= LargeLineCount {
                            fileJob.Content = nil
                            return
                        }
 
			switch currentState {
```

Worth it? Not sure. I don't know what amount of files that pass the byte limit that would hit this one to make it worth it. Trivial to implement though.

boyter: Added anyway.

boyter:
Oh one thing I should mention, when this gets in I can make a new release and then integrate scc into gitea which will be pretty cool!

boyter:
Done. Removed the additional stat call using the existing one already in both paths.

boyter:
Going to merge on the basis thats the last thing mentioned.

boyter:
Going to merge on the basis thats the last thing mentioned.

PR #131, opened 2019-10-25
Fix spelling of "Xcode"

dduan:
The "c" is lower case.


boyter:
Thanks for submitting. For future reference you need to run `go generate` to update the program itself assuming you were building from source. I have done so and pushed to master for you.

Issue #134, opened 2019-11-07
Processing files not in language list


ChristopherRogers:
**Describe the bug**
Filtering on a language properly filters out the output to only be files of that language, but it seems like there's an inefficiency where it's still processing or attempting to process all files regardless of this language filter.

**To Reproduce**

1. Run `scc -i language-of-choice --no-cocomo` in a directory & environment that will cause "too many files open" errors to be output.
1. Notice that the file names may be for files that aren't in the language filter passed. (Whether this actually occurs is likely dependent on the composition of the files.)

**Expected behavior**
Specifying the language of choice should result in scc not attempting to open files that aren't in the language list.

**Desktop (please complete the following information):**
macOS 10.15.1

boyter:
While that's not a great outcome I am more worried about the fact you are getting "too many files open". Certainly a performance optimisation in there potentially but the crash should not happen.

Are you able to share details about how many files are in the directory you are trying to count? Both file count and size of the directory. Id love to know how much RAM you have and what CPU you are running.

dbaggerman:
Google tells me the default open file limit on macOS is 256. We run `runtime.NumCPU() * 4` file readers, and `runtime.NumCPU()` gives the number of hyperthreads not physical cores.

According to apple.com, the current maximum spec for a Mac Pro is 28 cores, or 56 hyperthreads. `56 * 4 == 224`, which is getting fairly close to the limit, only leaving 32 spare handles. Might be worth capping `FileReadJobWorkers` at around 64/128, and/or printing a warning on high core count systems.

The directory walker also opens (directory) file handles, which also scale to the number of cores/hyperthreads. This would easily push through the 32 spare handles of the above scenario.

The directory handles can be kept open if the file job `chan` is full. We can close those directory handles a bit sooner if we change `defer file.Close()` to explicitly closing immediately after `file.Readdir()`. It won't reduce the upper limit of file handles, but it wouldn't be bad hygiene anyway.

ChristopherRogers:
I'm running this on an 18-core Xeon W (with hyperthreading) iMac Pro with 32 GB of RAM. The entire directory is 20.82 GB, with 81,899 files & directories. The files I'm interested in number 7,754 (per scc's output).

dbaggerman:
Based on that, I would expect an upper limit of 144 (36 * 4 file readers) + 36 (directory walkers), or 180 open files at a time. Reasonably high, but within the 256 limit that Google told me about. Maybe I'm missing something else that's keeping files open?

[...]

PR #135, opened 2019-11-08
Reduce dir handles

dbaggerman:
Relates to #134.

I found an issue in `cuba`, which allowed it to create more workers than it was supposed to. This meant that the `DirectoryWalker` could have more directory handles open than it was supposed to. Depending on the directory layout, it may have been significantly more.

This pulls in the a version (`v0.3.2`) with the fix.

While I was at it, I also refactored `DirectoryWalker` so that the handles get closed as soon as possible, rather than being kept open until after the entries have been processed.

### DIFF
--- a/processor/file.go
+++ b/processor/file.go
@@ -64,12 +64,12 @@ func NewDirectoryWalker(output chan<- *FileJob) *DirectoryWalker {
 		directoryWalker.excludes = append(directoryWalker.excludes, regexp.MustCompile(exclude))
 	}
 
-	directoryWalker.buffer = cuba.New(directoryWalker.Readdir, cuba.NewStack())
+	directoryWalker.buffer = cuba.New(directoryWalker.Walk, cuba.NewStack())
 
 	return directoryWalker
 }
 
-func (dw *DirectoryWalker) Walk(root string) error {
+func (dw *DirectoryWalker) Start(root string) error {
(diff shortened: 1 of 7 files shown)


processor/file.go
```diff
@@ -184,6 +177,21 @@ DIRENTS:
...
 }
 
+func (dw *DirectoryWalker) Readdir(path string) ([]os.FileInfo, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return []os.FileInfo{}, fmt.Errorf("failed to open %s: %v", path, err)
+	}
+	defer file.Close()
+
+	dirents, err := file.Readdir(-1)
```
boyter: Whats the purpose of -1? Can we just put a comment on what this actually does.

boyter:
Looks good. Only one question there.

processor/file.go
dbaggerman: From https://golang.org/pkg/os/#File.Readdir :

> If n <= 0, Readdir returns all the FileInfo from the directory in a single slice. In this case, if Readdir succeeds (reads all the way to the end of the directory), it returns the slice and a nil error. If it encounters an error before the end of the directory, Readdir returns the FileInfo read until that point and a non-nil error.

boyter: Ah cool. I was worried it was some magic number hack. Cheers.

PR #140, opened 2019-11-20
Issue 137

dbaggerman:
Refactors file reading and implements handling of `ulimit` on Unix-like operating systems.

For `ulimit` handling, it queries the current limits and raises the current (soft) limit if the hard limit allows it. If the hard limit is still too low then it will raise the current limit as high as it's allowed to and scale down the number of workers to fit.

Google is giving me mixed answers about what the default hard open file limit is, with everything from 512 to `unlimited`. I don't have a Mac to test with, but my Linux system has a soft limit of 2014 and a hard limit of 524288, so this will work fine for me until I can buy a CPU with 100,000 cores.

Testing this is awkward, but I ran `ulimit -n 50` in a shell and ran `scc` with `--verbose` to check the warning about scaling the limit. The hard limit was also set with `ulimit -Hn 50` to again see the warning message. In both cases I could also observe that `scc` didn't throw `too many open files` errors. I've got 8 cores / 16 threads, so I would normally expect to open up to 80 files at a time.

-----

I ended up folding the file read workers and file processor workers together. The advantage is that it lets us re-use the memory allocated for the file contents buffer, instead of allocating new memory for each file and having to garbage collect it. I explicitly drop the buffer if it's larger than `LargeByteCount` in case we end up reading any huge files.

For cases that are more CPU bound (i.e. all files are cached by the OS in memory), it benchmarks at ~8% faster for me.

For cases that are actually bound by disk IO, it's basically the same.

```
Benchmark #1: ./scc-master ~/src/linux                                                                                                                                                         
  Time (mean ± σ):     908.7 ms ±   8.9 ms    [User: 12.033 s, System: 1.171 s]
  Range (min … max):   891.5 ms … 949.1 ms    250 runs
 
Benchmark #2: ./scc-dev ~/src/linux
  Time (mean ± σ):     839.5 ms ±   7.5 ms    [User: 11.765 s, System: 0.743 s]
  Range (min … max):   820.9 ms … 868.7 ms    250 runs
 
Summary
  './scc-dev ~/src/linux' ran
    1.08 ± 0.01 times faster than './scc-master ~/src/linux'
```

### DIFF
--- a/processor/file_test.go
+++ b/processor/file_test.go
@@ -216,14 +216,7 @@ func TestNewFileJobFullname(t *testing.T) {
 	ProcessConstants()
 	AllowListExtensions = []string{}
 
-	fi, err := os.Stat("../examples/issue114/makefile")
-	if err != nil {
-		dir, _ := os.Getwd()
-		t.Logf("os.Getwd: %s", dir)
-
-		t.Errorf("os.Stat ./examples/issue114/makefile: %v", err)
-		t.FailNow()
-	}
+	fi, _ := os.Stat("../examples/issue114/makefile")
 	job := newFileJob("../examples/issue114/", "makefile", fi)
 
 	if job.PossibleLanguages[0] != "Makefile" {
--- a/processor/file_test.go
+++ b/processor/file_test.go
@@ -216,8 +216,15 @@ func TestNewFileJobFullname(t *testing.T) {
 	ProcessConstants()
 	AllowListExtensions = []string{}
 
-	fi, _ := os.Stat("./examples/issue114/makefile")
-	job := newFileJob("./examples/issue114/", "makefile", fi)
+	fi, err := os.Stat("../examples/issue114/makefile")
+	if err != nil {
+		dir, _ := os.Getwd()
+		t.Logf("os.Getwd: %s", dir)
+
+		t.Errorf("os.Stat ./examples/issue114/makefile: %v", err)
+		t.FailNow()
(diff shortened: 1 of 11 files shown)


processor/file_test.go
```diff
@@ -227,8 +234,8 @@ func TestNewFileJobFullname(t *testing.T) {
 func TestNewFileJob(t *testing.T) {
 	ProcessConstants()
 
-	fi, _ := os.Stat("./examples/issue114/java")
-	job := newFileJob("./examples/issue114/", "java", fi)
+	fi, _ := os.Stat("../examples/issue114/java")
+	job := newFileJob("../examples/issue114/", "java", fi)
```
dbaggerman: I'm not sure why I had to change these. Has something about the testing changed?

boyter: No... that's odd... 

How were you running them?

dbaggerman: Both `go test ./...` and `sh test-all.sh` - the same way I've run the tests in the past.

boyter:
Looks good to me. Merge whenever you feel comfortable.

dbaggerman:
> Looks good to me. Merge whenever you feel comfortable.

I've been meaning to take another look at where the test coverage went. But I suppose I'll merge as is and put another PR up if I tweak the tests.

boyter:
Yep sounds fine. I tend to have a bash at improving them every now and then as well.

Issue #139, opened 2019-11-14
Add a HTML output formation option


eric-zheng:
**Describe the bug**
When I run scc on command line, the report table shows correctly on the command line terminal. But if I pipe the result to a file and open the file with common editor / browser, the line separator of the table is displayed incorrectly. I tried to change the character encoding but couldn't fix the issue.

**To Reproduce**

```bash
.\scc.exe --version
scc version 2.10.1

.\scc.exe > report.txt
.\scc.exe > report.html
```

Then, open the report.txt with code editor and open the report.html with browsers (I used Chrome and Edge). The line separator is encoded as `ΓöÇΓöÇΓöÇΓöÇ`.

**Expected behavior**

I think the line separator of the table is not the common ascii "-" dash character. But I'm not sure how to encode it correctly on browser/editor. It would be good if scc supports to generate a file output with certain extensions (.html, .txt) that display the report table correctly. Thanks!

**Desktop (please complete the following information):**
 - OS: Windows, Linux
 - Version: 2.10.1

boyter:
Thankfully there is already an option to deal with this.

`scc --ci`

Which will output the format using - and not box lines which should resolve this issue for you. In effect it reduces the output to pure ascii and not unicode.

I suspect based on your name that when you redirect the output you are not specifying the encoding type, and as such it the encoding is incorrect from your terminal to your browser resulting in the symptom.

Can you please try with the `--ci` option which should resolve this in the short term. In the long term if you wish to see box lines you should confirm that your terminal and browser are using the same encoding type.

boyter:
A HTML output option might be worth implementing. Probably just need to output to a simple table with headers. Should make the enclosing of a <html></html> an optional param to turn the output into a self contained format. Could possibly link out to https://www.datatables.net/ to provide a nice interface, but that seems like overkill.

Might be worth looking to see if any of the other tools do this as well.

eric-zheng:
Appreciate that you would like to consider this feature. I will check tomorrow to see if `--ci` works for me. But generally speaking, an HTML option would be helpful. Thanks for making this awesome (fast) tool.

boyter:
Have a PR going with this implemented. Example screenshot of the `html` output (full page)

![image](https://user-images.githubusercontent.com/612151/71789515-8f37cc80-307f-11ea-9963-3858e2bd3108.png)

and there is an option for just the table output if you want to embed into another page.

boyter:
[...]

PR #146, opened 2020-01-06
Html output 139

boyter:
Adds HTML output as an output option. I don't see too many people wanting this but it has been requested https://github.com/boyter/scc/issues/139

Supports all of the options that the command line does, except for the case of omitting complexity, although if you add that it sets it to 0 for everything.

Two modes, `html` and `html-table` where the former includes html and body tags IE a single HTML file ready to view (see screenshot) and the other which is designed to be embedded into another page and the user brings their own CSS.

image

### DIFF
--- a/processor/formatters.go
+++ b/processor/formatters.go
@@ -283,6 +283,10 @@ func toCSV(input chan *FileJob) string {
 	return b.String()
 }
 
+func toHtml(input chan *FileJob) string {
+	return `<html lang="en"><head><meta charset="utf-8" /><title>scc html output</title></head><body>` + toHtmlTable(input) + `</body></html>`
+}
+
 func toHtmlTable(input chan *FileJob) string {
 	languages := map[string]LanguageSummary{}
 	var sumFiles, sumLines, sumCode, sumComment, sumBlank, sumComplexity int64 = 0, 0, 0, 0, 0, 0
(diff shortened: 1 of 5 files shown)


processor/formatters.go
```diff
@@ -284,6 +283,122 @@ func toCSV(input chan *FileJob) string {
...
+			}
+		}
+	}
+
+	language := []LanguageSummary{}
+	for _, summary := range languages {
+		language = append(language, summary)
+	}
+
+	language = sortLanguageSummary(language)
```
dbaggerman: Looks like this code to summarise the results is duplicated across multiple formatters - maybe it can be lifted out to a shared function?

processor/formatters_test.go
```diff
@@ -30,6 +30,18 @@ func TestPrintError(t *testing.T) {
...
+}
+
+func TestPrintDebugF(t *testing.T) {
+	printDebugf("Testing print error")
+}
+
+func TestPrintTraceF(t *testing.T) {
+	printTracef("Testing print error")
+}
+
```
dbaggerman: I reserve the right to point this out next time you say that you're not that pedantic about test coverage ;)

dbaggerman:
Other than maybe a bit of duplication, LGTM.

boyter:
Yeah was planning to refactor it out later :) ill merge for the moment and work on it a bit later.

processor/formatters_test.go
boyter: :P I thought it was an easy gain :) and since its displayed on the main page...

processor/formatters.go
boyter: Yep it is. Ill merge then refactor it a bit later.

Issue #128, opened 2019-10-20
Enhancement: Parse As


boyter:
Would be neat to be able to say "this extension parse as C" or some such. Would unblock those who need `scc` to parse a specific type which is not yet included, or if they are using a custom extension for some reason.

No idea how this would work from command line but something like, the below to change the count rules to count JSP using HTML rules

```
scc --count-as=jsp,html
```
However more useful would be something like

```
scc --count-as=boyter,java
```

Where the new extension boyter is counted as java

Would probably need to allowlist the boyter. Although it might just be a case of taking the Java rules and extensions and updating for the new logic after running `ProcessConstants`.

PR #147, opened 2020-01-07
Parse as 128

boyter:
Add functionality to overwrite extensions https://github.com/boyter/scc/issues/128

### DIFF
--- a/main.go
+++ b/main.go
@@ -212,7 +212,7 @@ func main() {
 		&processor.CountAs,
 		"count-as",
 		"",
-		"count extension as language [e.g. jsp,htm:chead,\"C Header\" maps extension jsp to html and chead to C Header]",
+		"count extension as language [e.g. jsp:htm,chead:\"C Header\" maps extension jsp to html and chead to C Header]",
 	)
 
 	if err := rootCmd.Execute(); err != nil {
--- a/main.go
+++ b/main.go
@@ -212,7 +212,7 @@ func main() {
 		&processor.CountAs,
 		"count-as",
 		"",
-		"count extension as another [e.g. jsp,html:new,java counts JSP using HTML rules and new using Java rules]",
+		"count extension as language [e.g. jsp,htm:chead,\"C Header\" maps extension jsp to html and chead to C Header]",
 	)
 
 	if err := rootCmd.Execute(); err != nil {
--- a/main.go
+++ b/main.go
@@ -208,6 +208,12 @@ func main() {
 		1000000,
 		"number of bytes a file can contain before being removed from output",
 	)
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -208,6 +208,12 @@ func main() {
 		1000000,
 		"number of bytes a file can contain before being removed from output",
 	)
+	flags.StringVar(
+		&processor.CountAs,
+		"count-as",
+		"",
+		"count extension as language [e.g. jsp,htm:chead,\"C Header\" maps extension jsp to html and chead to C Header]",
```
dbaggerman: Purely personal preference, but using `:` to separate key/value and `,` to separate the pairs seems more "normal" to me. My brain keeps wanting to associate `htm` with `chead` in that example.

i.e. `jsp:htm,chead:"C Header"`

boyter: Yeah I was not sure which way to go with it. Ill flip it around. I don't have a strong preference either way.

dbaggerman: My actual preference would be to have one flag per re-mapping (e.g. `--count-as jsp:htm --count-as chead:"C Header"`), but that seems rather awkward to do with just the basic `flags` package.

boyter: Yeah one of those annoying things. Something to make as a breaking change in the future perhaps.

processor/processor.go
```diff
@@ -191,7 +199,55 @@ func ProcessConstants() {
...
+					}
+				}
+			}
+
+			// If the above did not work, its a matter of extension match
+			// note that this is less reliable as some languages share extensions
+			if !identified {
+				for name, value := range languageDatabase {
+					for _, ext := range value.Extensions {
+						if strings.ToLower(ext) == strings.ToLower(t[1]) {
```
dbaggerman: Since we're doing this after building `ExtensionToLanguage`, we should be able to look up from that rather than iterating through the entire language list for every remapping.

boyter: Good point. I moved it out of the previous loop because it would have made it do it every time. Didn't even think to use the lookup itself.

processor/processor.go
```diff
@@ -230,20 +230,22 @@ func setupCountAs() {
 						printDebug(fmt.Sprintf("set to count extension: %s as language %s by language", t[0], name))
 					}
 				}
+
+				if !identified {
+
+				}
```
dbaggerman: This looks like you committed  something by accident.

dbaggerman:
Other than the empty condition, looks good to me.

processor/processor.go
boyter: How did I manage that... oh well.

PR #151, opened 2020-01-20
Detect generated files by heading comment

vearutop:
By convention machine-generated `Go` files have heading comment in particular format.
```
 ^// Code generated .* DO NOT EDIT\.$
```
https://github.com/golang/go/issues/13560#issuecomment-288457920

This PR adds a check to detect machine-generated files by having case insensitive `generated` or `do not edit` in comments of first 1000 bytes, so that such files can be grouped together when `-z` is specified.

```
scc -z
Language                 Files     Lines   Blanks  Comments     Code Complexity
Go                          82      4909      881       325     3703        390
Java                        15       525      103         0      422          3
Go (gen)                    14      1715      362       144     1209        275
YAML                        13      1836      211       182     1443          0
```


processor/workers.go
```diff
@@ -507,6 +510,12 @@ func CountStats(fileJob *FileJob) {
 			case SComment, SMulticomment, SMulticommentBlank:
 				fileJob.Comment++
 				currentState = resetState(currentState)
+				if MinifiedGenerated && !generatedComment && lineStart < 1000 {
```
boyter: The problem I see with doing this here is that it will run over and over again for the first 1000 bytes which is going to impact performance. This logic should be pulled out above the core loop as it only needs to be done once.

That's going to be tricker though as you will need to handle the newline + bounds check yourself and ensure there is a comment. It might be acceptable to drop the comment check though as if you see the text generated or do not edit that might be good enough.

processor/workers.go
```diff
@@ -527,6 +536,12 @@ func CountStats(fileJob *FileJob) {
 				}
 			case SDocString:
 				fileJob.Comment++
+				if MinifiedGenerated && !generatedComment && lineStart < 1000 {
```
boyter: Same as above this should not be in the core loop.

processor/workers.go
```diff
@@ -545,12 +562,19 @@ func CountStats(fileJob *FileJob) {
...
-	if avgLineByteCount >= MinifiedGeneratedLineByteLength {
+func minifiedGeneratedCheck(avgLineByteCount int, fileJob *FileJob, generatedComment bool) {
+	if generatedComment {
+		fileJob.Minified = true
+		fileJob.Language = fileJob.Language + " (gen)"
+
+		if Verbose {
+			printWarn(fmt.Sprintf("%s identified as generated with heading comment", fileJob.Filename))
+		}
+	} else if avgLineByteCount >= MinifiedGeneratedLineByteLength {
```
boyter: We saying if its generated its no minified here. 

I think that might be correct. But want to check with @dbaggerman for thoughts

boyter:
My overall comment is maybe calling it minfied/generated was wrong. Perhaps they should be split and separate options. So one flag for minified and another for generated. That way you could get both or individual.

processor/workers.go
vearutop: Also I think it is better to remove `generated` from this check, and only keep `do not edit`, as `generated` alone produces false positives (even in `scc` repo).

boyter: I was wondering about that. False positives are not ideal.

I think adding an option to supply the string to search for would help with this as well. So you invoke scc like so

```
scc --generated-string "do not edit",generated
```

or some such. This forces the user to implement it themselves avoiding the chance of induced false positives. 

Although a `-g` option or some such with some known generated types might still be useful as a base case.

boyter:
So I like this as a feature. Happy to support it long term assuming we get the details correct. I think we should do the following.

1. Move the check outside of the main loop as mentioned above, for performance.
2. Change the option to be separate from the minified check, with the option being `--generated-string` where the user supplies the generated strings EG `--generated-string "do not edit",generated` which are scanned for.
3. A potential idea is have a default option --gen which looks for a list of generated strings. Will need to do some looking around to find out what those strings should be to try and get them to be reasonably accurate out of the box.
4. Would also need a `--no-generated` option to exclude the generated files from the output similar to minified.

Annoyingly this means the --min-gen seems a bit wrong BUT as minified files are usually generated I think we can live with it. Would just need to change its message where it say generated but that's not a huge issue. 

Just my thoughts at the moment.

boyter:
Just had a better thought regarding the spit of arguments. 

Leave `--min-gen` there as an option that turns on both checks for minified and generated files and create new ones just for minified and generated. Keeps backwards compatibility and provides a shortcut if you want both.

vearutop:
@boyter could you suggest a strategy to check head of file for marker (`do not edit`)? 

Since performance is important, is it acceptable to have something like `strings.Contains(strings.ToLower(string(fileJob.Content[0:1000])), "do not edit")` or should I implement something more complicated to minimize casting/lowering penalty?

boyter:
I think that is acceptable as it would only happen if the user opts in so by default performance is not an issue and the user is saying they are willing to pay the penalty this incurs.

vearutop:
@boyter I've updated this PR to meet the requirements discussed in comments, please check if I missed or misunderstood something. :bowing_man:

vearutop:
On a side note, I thought it would be also helpful to implement separate test code counter for Go (`_test.go`), if it makes sense I can send a PR. Not sure how feasible that would be for other languages.

boyter:
Cheers. Will have a look over the next few days. Cursory glace looks good though.

As for separate test counter, sure. I don't mind so long as the tests are there. Splitting them out to separate files might be a good idea in the long run.

boyter:
Looks good to me. Merging in. Thanks so much for this submission @vearutop and I will ensure you get credit in the release notes.

PR #157, opened 2020-02-04
Add terraform language support

lingrino:
This PR adds support for [Terraform](https://www.terraform.io/). Let me know if there's anything to change, some of the `complexitychecks` are unconventional and I'm not sure what exactly they should be.

`go test ./...` is passing but not `./test-all.sh`, which is also failing for me on master.

Terraform [Syntax](https://www.terraform.io/docs/configuration/syntax.html) and [Expression](https://www.terraform.io/docs/configuration/expressions.html) Docs


boyter:
Looks good to me. Thanks for the nicely done PR.

PR #160, opened 2020-02-13
Add cljc file extension for Clojure

jimmyhmiller:
cljc files allow you to write code that is [portable between clojure dialects](https://clojure.org/guides/reader_conditionals). Many clojure projects use these files and not having them makes scc greatly under count how many lines a project has.

I explicitly licence my contribution under MIT AND The Unlicense.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1160,7 +1160,8 @@
       "== "
     ],
     "extensions": [
-      "clj"
+      "clj",
+      "cljc"
     ],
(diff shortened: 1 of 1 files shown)


boyter:
Looks good to me. Will merge and regenerate the constants so it picks up. You can then build from master or wait for the next release to get this.

Also thank you for explicitly telling me how you licensed this. Makes it much easier.

Issue #169, opened 2020-04-23
move to go modules


5paceToast:
**Describe the bug**
You cannot clone the repository and run `go build .` outside of GOPATH.

**To Reproduce**
1. git clone https://github.com/boyter/scc.git
2. cd scc
3. go build .

**Expected behavior**
SCC builds.

**Desktop (please complete the following information):**
Arch Linux

**Recommended approach:**
1. run `go mod init github.com/boyter/scc` - this will parse the current system (gopkg) and replicate
2. regenerate vendor directory (or remove vendoring)

boyter:
I guess `go modules` are mature enough now to try it out.

5paceToast:
They're actually fully released as of go 1.14.
To quote the [release blog post](https://blog.golang.org/go1.14):
> Module support in the go command is now ready for production use. We encourage all users to migrate to go modules for dependency management.

I've been using them since introduction and haven't had any issues though.

boyter:
I ran into a lot of issues when I first tried it. I did not it was prod ready in 1.14 and thought I should give it another go. Will try it again soon.

boyter:
Graciously done by @dbaggerman

PR #170, opened 2020-04-28
Add .erb extension for "Ruby HTML"

dbackeus:
The `.erb` extension has been used over `.rhtml` in Ruby on Rails for over a decade at this point.

Btw I've never worked with go and haven't figured out how to actually build the project (possibly in part due to https://github.com/boyter/scc/issues/169) so have not been able to verify if this is working as intended.


boyter:
Looks like you did it anyway. Cheers for the submission.

PR #171, opened 2020-04-28
Go module conversion

dbaggerman:
Relates to  #169 - I've been using go modules exclusively for a while. I think `scc` is the only project I still need to maintain my `GOPATH` for. This is the result of `go mod init` and `go mod vendor`.


boyter:
Yeah just me being lazy having left it like this. Looks good to me.

boyter:
GAH that stupid gitignore issue is back... ill patch it back in now.

PR #201, opened 2020-09-18
Added ReasonML support

happylinks:
https://reasonml.github.io/

Mostly copied the JavaScript implementation because it's comparable.
I wasn't sure about `shebangs` and couldn't find immediately what that should be set as.

Let me know if I should change something :)


languages.json
```diff
@@ -7124,5 +7124,42 @@
...
+      ]
+    ],
+    "quotes": [
+      {
+        "end": "\"",
+        "start": "\""
+      }
+    ],
+    "shebangs": [
+      "node"
```
boyter, line 7162: Need to remove this :) 

It will conflict with the existing one. Not sure if there is some specific shebang for reason.

boyter:
Only issue I can see is the shebang. It will conflict with the existing Javascript ones, which is probably not what we want. Is there some specific shebang for reason?

Otherwise it looks good!

happylinks:
Just pushed it without `shebangs`, I don't think Reason has them specifically, since it's compiled to either OCaml or Javascript.
Let me know if I have to change anything else :)

boyter:
Looks good to me!

PR #205, opened 2020-11-12
Added support for remapping extensions

davidribyrne:
Added a --RemapExtensions option that allows users to force extensions to be mapped to a specific language.

### DIFF
--- a/main.go
+++ b/main.go
@@ -289,6 +289,12 @@ func main() {
 		"",
 		"inspect every file and remap by checking for a string and remapping the language [e.g. \"-*- C++ -*-\":\"C Header\"]",
 	)
+	flags.StringSliceVar(
+		&processor.RemapExtensions,
+		"remap-extensions",
+		[]string{},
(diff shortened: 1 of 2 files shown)


boyter:
I may be missing something here, but how is this different to the currently implemented

```
--count-as string             count extension as language [e.g. jsp:htm,chead:"C Header" maps extension jsp to html and chead to C Header]
```

There is probably something I am missing here that you are aware of.

davidribyrne:
🤯 Well, that's embarrassing; I don't know how I missed that flag. On the positive side, I became more familiar with Go and with your code (which is well organized). Thank you for your nice work.

davidribyrne:
Thank you for the fast response too

boyter:
Hah no hard feelings at all. I thought maybe there was something I was missing. Glad you liked the code and I do appreciate the submission.

I wonder if I could change the README to make this functionality more obvious somehow to avoid people missing it such as yourself.

PR #209, opened 2020-12-11
SLOCCount came before cloc

AlDanial:
hi, great project; just want to correct some history in the README--I wrote cloc after SLOCCount had already been around a few years.


boyter:
TIL. Merged. Cheers.

PR #217, opened 2021-01-04
Add "gohtml" and "gotxt" to the list of extensions for Go templates

arp242:
These are the default extensions that e.g. GoLand uses, and some other
tools as well I believe. It's useful as it disambiguates between
templates for html/template and text/template.


boyter:
Looks good! Thanks for the submission and it will be in the next release.

Issue #214, opened 2020-12-23
Long filename contains unicode characters with `by-file` truncates error


lingsamuel:
**Describe the bug**
Long filename contains unicode characters with `by-file` truncated incorrectly.

**To Reproduce**

```bash
touch "ббббббббббббббббббббббббббббббббб.md"
touch "ѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬѬ.md"
touch "中文中文中文中文中文中文中文中文中文中文中文中文中文中文中文中文.md"
scc --by-file
```

Output:

![image](https://user-images.githubusercontent.com/14567045/102988542-01cee000-454f-11eb-882d-fab28097b734.png)

**Expected behavior**
`��` shouldn't appear, indent should works properly.

**Desktop (please complete the following information):**
 - OS: Linux 5.9.9-arch1-1 (Arch Linux)
 - scc version 2.13.0, make from source

lingsamuel:
б https://www.compart.com/en/unicode/U+0431
Ѭ https://www.compart.com/en/unicode/U+046C
中 https://www.compart.com/en/unicode/U+4E2D

boyter:
You would think I would be aware of this considering how much time I have spent looking over unicode recently.

Its down to it using bytes and not characters. Its especially annoying because it trims from the front, which isn't how Go lets you walk runes.... but not impossible to rectify. I can see the size never matching up though depending on the characters though, but we can fix the current issues for sure.

boyter:
Done. Note there isn't much I can do about 中 which for some reason does not display in mono-spaced font... tokei has the same issue. However at least its not cut off anymore when printing,

![image](https://user-images.githubusercontent.com/612151/103590274-83f0d880-4f41-11eb-9fed-fe0b5f8479ff.png)

dbaggerman:
> Note there isn't much I can do about 中 which for some reason does not display in mono-spaced font

You could have a play around with https://godoc.org/golang.org/x/text/width to calculate the display width.

boyter:
I actually found this just then http://github.com/mattn/go-runewidth which might be able to solve this.

boyter:
Added the above in... and its almost there.. however it appears sprintf has issues with this as well... but moved over to some custom parsing. Anyway merged in to master.

[...]

PR #218, opened 2021-01-04
partial fix for https://github.com/boyter/scc/issues/214

boyter:


processor/formatters.go
```diff
@@ -750,6 +744,24 @@ func fileSummarizeLong(input chan *FileJob) string {
...
 
+// We need to trim the file display for tabular output formats which this does in a unicode aware way
+// to avoid cutting bytes... note that it needs to be expanded to deal with longer display characters at some
+// point in the future
+func unicodeAwareTrim(tmp string, size int) string {
+	// iterate all the runes so we can cut off correctly and get the correct length
+	var r []rune
+	for _, x := range tmp {
+		r = append(r, x)
+	}
```
dbaggerman, line 755: FYI, you can cast `[]rune(tmp)` the same way you would to a `[]byte`.

boyter: Ah neat. Updated.

PR #222, opened 2021-01-20
now in scoop!

batkiz:
Check https://github.com/ScoopInstaller/Main/blob/master/bucket/scc.json

sorry for my latency


boyter:
Woah neat! Cheers!

PR #231, opened 2021-02-21
Add DM language support

mloc:
Language ref: http://www.byond.com/docs/ref/


boyter:
Nice! Thanks for the excellent PR.

PR #232, opened 2021-02-22
Issue164

boyter:
Very small performance tweak on some workloads. For languages identified as "simple" which means there is nothing to count, flip into simple mode and skip all of the logic of the state engine. This is mostly in preparation of having custom state machines for more common languages which will follow the same ideas hence being a bit ugly.

```
$ hyperfine -m 250 'scc' './scc'
Benchmark #1: scc
  Time (mean ± σ):      37.9 ms ±   3.1 ms    [User: 111.0 ms, System: 78.8 ms]
  Range (min … max):    31.0 ms …  43.5 ms    250 runs
 
Benchmark #2: ./scc
  Time (mean ± σ):      37.0 ms ±   3.7 ms    [User: 111.5 ms, System: 78.3 ms]
  Range (min … max):    30.3 ms …  60.8 ms    250 runs
 
Summary
  './scc' ran
    1.02 ± 0.13 times faster than 'scc'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -705,13 +705,13 @@ scc is pretty well tested with many unit, integration and benchmarks to ensure t
 Run go build for windows and linux then the following in linux, keep in mind need to update the version
 
 ```
-GOOS=darwin GOARCH=amd64 go build -ldflags="-s -w" && zip -r9 scc-2.13.0-x86_64-apple-darwin.zip scc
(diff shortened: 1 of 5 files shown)


boyter:
Need to keep evaluating this to see if it really is an improvement.

boyter:
Actually this does not account for blank lines which needs to be in there. 

Hmmm.

PR #236, opened 2021-02-27
Add TypeLanguage support

ernado:
Ref:
https://core.telegram.org/mtproto/TL


boyter:
Thanks for the submission!

Issue #246, opened 2021-03-28
Python docstrings mishandle apostrophes 


dbaggerman:
**Describe the bug**
Python docstrings containing an apostrophe end up in a weird state, continuing the docstring parsing beyond the end of the docstring and then ending up in a string state waiting for a closing single quote / apostrophe.

**To Reproduce**

```python
#!/usr/bin/env python3

"""
Docstrings containing an apostrophe (') are handled incorrectly
The line above is counted as code despite being in the middle of a docstring.
The end of docstring flag seems to be changed to an apostrophe,
which means the next line will not exit the docstring.
"""
# Code containing single quotes will exit the docstring,
# but presuming the quotes are balanced the second
# quote will put us in string scanning mode.
if __name__ == '__main__':
    print('Hello, World!')
# Not counted as a comment

# ^ Not counted as a blank line
# Break out of string scanner with unbalanced single quote: '
    exit(0)
```

**Expected behavior**
Apostrophes / single quotes inside a docstring don't interrupt docstring parsing.

**Desktop (please complete the following information):**
 - Version 3.0

PR #247, opened 2021-03-28
Refactor CountStats state machine

dbaggerman:
Since I've been back in the `scc` code, I got tempted to make an(other) attempt to break up the complexity in the `CountStats` state machine.

This changes from states being an enum, with all the state tracked in `CountStats` and passed in/out of the different methods to introduce a `State` interface with different states having their `stateXYZ` function moved to being attached to the relevant struct. This means we can drop the `switch` statement in the core loop and instead let the go runtime do the dispatch. It also lets relevant state be tracked in the `State` struct instead of being passed in and out of `CountStats`.

As a bonus, the different 'flavours' of `slComment` vs `slCommentCode`, etc are no longer required.

It seems to produce the same result in the examples I've thrown at it, other than python docstrings - issue #246 doesn't exist in this implementation (this mismatch is how I found that issue in the first place). It could probably use a bit more hammering just to make sure though - I haven't tested permutations of optional flags for example.

Performance is roughly on par with the current implementation, only being marginally slower:
```
Benchmark #1: ./scc-master ../../torvalds/linux
  Time (mean ± σ):     688.1 ms ±  11.9 ms    [User: 8.613 s, System: 0.970 s]
  Range (min … max):   663.2 ms … 712.3 ms    50 runs

Benchmark #2: ./scc-refactor ../../torvalds/linux
  Time (mean ± σ):     695.7 ms ±  10.4 ms    [User: 8.567 s, System: 1.034 s]
  Range (min … max):   675.1 ms … 720.2 ms    50 runs

Summary
  './scc-master ../../torvalds/linux' ran
    1.01 ± 0.02 times faster than './scc-refactor ../../torvalds/linux'
```

### DIFF
--- a/examples/issue246.py
+++ b/examples/issue246.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+"""
+Docstrings containing an apostrophe (') are handled incorrectly
+The line above is counted as code despite being in the middle of a docstring.
+The end of docstring flag seems to be changed to an apostrophe,
+which means the next line will not exit the docstring.
+"""
+# Code containing single quotes will exit the docstring,
+# but presuming the quotes are balanced the second
+# quote will put us in string scanning mode.
+if __name__ == '__main__':
+    print('Hello, World!')
+# Not counted as a comment
+
+# ^ Not counted as a blank line
+# Break out of string scanner with unbalanced single quote: '
+    exit(0)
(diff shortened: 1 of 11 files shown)


boyter (approved):

boyter:
Yeah the state machine stuff is something I have been toying around with as well. I want too start pulling out common languages into their own versions which should allow for further performance optimisations. For example, counting text files should be very fast as they only need to check for new lines, and if the line is all whitespace to produce the correct counts. 

Something ill get back into at some point.

PR #248, opened 2021-04-03
Fix vimscript comments and files

LunarWatcher:
`.vimrc` and `.gvimrc` (along with their global variants and windows-specific names) were ignored. Comments were also broken (searched for `\"` instead of `"`).

Also added support for vim9 comments (see: https://vim-jp.org/vimdoc-en/vim9.html#vim9-differences)

And a general comment: the general language config is extremely inconsistent between `\\\"` and `\"` when normal quotes are listed. That being said, I had to leave the quote config to search `\"` instead of `"`. Vimscript uses a single `"` to mean comments, but `""` on a single line can refer to a string instead. Also means this is valid: `let test = "This is a string" " But this is a comment`

scc doesn't seem to account for context with languages like these, so it's an either-or situation without a patch for that (which I'm not competent enough in Go to write). Not sure what string counting is for (might be for ignoring comments in multiline strings?), but comment counting at least works now.

Anyway, before (`.vimrc`, not `test.vim`):

image

And before with my `.vimrc` temporarily copied to `test.vim`:

image

After (with `.vimrc` and not `test.vim`):

image

And finally, source for the filenames:

image

(`_vimrc` and `_gvimrc` are windows files, which aren't listed because I'm not on windows. Their validity can be verified with `:version` in vim on Windows. `.exrc` is a Vi file and not directly vim, so decided to exclude that)


boyter:
Looks good thanks for the addition and the details above.

Having context for the above is something I think best solved by a specific parser for this language, and others as well as a part of getting additional performance out of the tool as well. Have created this https://github.com/boyter/scc/issues/249 in order to keep track of it.

Issue #251, opened 2021-04-18
the way to identify escape character in state_string.go need to be improved


foxdd:
**Describe the bug**
it's in the 25th line of state_string.go. Currently, only fileJob.Content[i-1] is put into consideration that if it is back slash. But what if two consecutive back slashes there which represents one escape bachslash character and has nothing to do with the following end string? Therefore, four, six... back slashes are also possible.

**To Reproduce**

1. As it is described ahead

**Expected behavior**
we need to judge the number of back slashes before the end string are even, so we can process the next checkForMatchSingle function

**Desktop (please complete the following information):**
 - OS: [e.g. macOS, Windows, Linux]

foxdd:
see pull request #252

boyter:
This code has been reverted due to all of the issues so I am going to close this down.

PR #252, opened 2021-04-18
correct the escape character judgement for string state

foxdd:


foxdd:
to solve the issue #251

boyter (approved):

boyter:
Thanks for both the bug report and the fix!

PR #254, opened 2021-04-22
Make the output file permission from 600 to 644 for other user reading.

fredgan:
Make the output file permission from 600 to 644 for other user reading.


boyter:
Ah good point. Probably makes sense in the scheme of CI/CD tools. Thanks for the PR.

Issue #256, opened 2021-04-26
Python Complexity Checks


TomFryers:
In `languages.json`, the Python `complexitychecks` are:
- `for `
- `for(`
- `if `
- `if(`
- `switch `
- `while `
- `else `
- `and `
- `or `
- `not `
- `in `
These don’t seem right. I think (although some of these may be wrong: I’m not sure exactly how `scc` uses them)
- `switch` should be `match` (which arrives in Python 3.10),
- `else ` should be `else:`,
- `elif ` (and perhaps `elif(`) should be there,
- maybe `try:`, `except ` and `finally:` should be added, since these are sort of branching,
- `in ` maybe shouldn’t be there (it might mean a for loop (`for item in iterable:`) gets counted double), and
- `== ` and `!= ` should exist, like they do for C and Java (although I don’t know why they are there for these languages). Maybe the other comparison operators (`< `, `<= `, `> `, `>= `, `is ` (and `in `)) should also be there?

boyter:
I think I just copied from Java or something when setting this up. 

Agreed they should be changed to Python like ones. If you want to do it and submit a PR you can have the credit :) Its pretty simple, just update languages.json and you are done.

BrandonKMLee:
So Go, Python, Java, C++, Kotlin, JS all have different complexity tokens...

boyter:
Yes they should. When I set them up I had a quick look at the language descriptions, specifically the statements that can cause a cpu branch condition and set those up. The are not directly comparable unless the languages themselves have close to identical complexity conditions.

boyter:
Resolved in master now.

PR #257, opened 2021-04-27
Fix Python complexity checks

TomFryers:
I’ve also put in `with` because it’s a shortcut to `try` `except` `finally` blocks, and made a space and open bracket/colon version of each. `keyword(` and `keyword :` are against Python style, but not grammar.


boyter:
Neat. Thanks for doing this. 

Note to self, relates to https://github.com/boyter/scc/issues/256

PR #262, opened 2021-05-25
Add initial support for Dyalog APL

stampes:
Add support for Dyalog APL file formats.


boyter:
Nice. Thanks for the contribution.

PR #272, opened 2021-07-17
use Compile instead of MustCompile to avoid panic

### LINKED ISSUE #273: Suggestion: use 'Compile' function to get regular expression of exclude path
we've met a problem in the daily use of SCC tool, which is, when the input exclude path list contains an incorrect regular expression, the entire scan process will be interrupted and a panic is thown out. So i think Compile function may be more user-friendly, so the process will continue and only the incorrect exclude path will be ignored

foxdd:
Fixes #273


processor/file.go
```diff
@@ -66,7 +66,10 @@ func NewDirectoryWalker(output chan<- *FileJob) *DirectoryWalker {
 		output: output,
 	}
 	for _, exclude := range Exclude {
-		directoryWalker.excludes = append(directoryWalker.excludes, regexp.MustCompile(exclude))
+		regexpResult, err := regexp.Compile(exclude)
+		if err == nil {
+			directoryWalker.excludes = append(directoryWalker.excludes, regexpResult)
+		}
```
boyter, line 72: Can you put a condition on there to handle if there is an error,

```
} else {
    printError(err.Error())
}
```

Then at least its explicitly listed why it didn't work as expected, but without interrupting the run itself.

boyter (changes requested):

processor/file.go
foxdd: sure, code updated, plz check pr again

boyter (approved):

Issue #281, opened 2021-08-19
(Maybe) Incorrect cost in COCOMO calculations


Theldus:
## Description & steps to reproduce
Hi,
I've been testing `scc` on a project of mine (PBD) and the estimated cost just seems to be way higher than expected:
```bash
$ scc --version
scc version 3.0.0

$ git clone https://github.com/Theldus/PBD
Cloning into 'PBD'...
remote: Enumerating objects: 845, done.
remote: Counting objects: 100% (493/493), done.
remote: Compressing objects: 100% (304/304), done.
remote: Total 845 (delta 327), reused 318 (delta 164), pack-reused 352
Receiving objects: 100% (845/845), 248.96 KiB | 0 bytes/s, done.
Resolving deltas: 100% (590/590), done.

$ scc PBD/
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
C                           16      7856      927      2645     4284        796
C Header                    15      1543      186       622      735        113
Makefile                     3       303       42       115      146          2
Assembly                     2       502       38         0      464          8
Shell                        2       252       31        79      142         12
License                      1        19        3         0       16          0
Markdown                     1       223       44         0      179          0
R                            1        98       11        35       52          0
YAML                         1        49        4         4       41          0
gitignore                    1        33        1        22       10          0
───────────────────────────────────────────────────────────────────────────────
Total                       43     10878     1287      3522     6069        931
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop (organic) $179,418
Estimated Schedule Effort (organic) 7.159522 months
Estimated People Required (organic) 2.226385
───────────────────────────────────────────────────────────────────────────────
Processed 292631 bytes, 0.293 megabytes (SI)
───────────────────────────────────────────────────────────────────────────────
```
I may be completely equivocated (and I apologize in advance), but 179k$ for a 6096 SLOC project, 2 developers and 7 months, seems _far_ above the average developer salary.

## Initial investigation
Although I'm not a Go programmer, below is my attempt to understand better the issue.

The organic COCOMO calculation code ([`cocomo.go`](https://github.com/boyter/scc/blob/master/processor/cocomo.go)) can be quickly implemented with the following Python code:
```python
def estimate_cost(effort, wage):
    return effort * float(wage/12) * 2.4

def estimate_effort(sloc):
[...]

PR #298, opened 2021-10-29
Add a overhead commandline flag to set the multiplier for corporate overhead. #281

fschaefer:
This commit makes the hard-coded overhead factor of 2.4 configurable by command line argument. Issue #281 is maybe related.


boyter:
Nice!

PR #299, opened 2021-11-01
Add a command line argument for the effort adjustment factor.

fschaefer:
Here is another one for the effort adjustment factor (EAF). The EAF can be used to set deviating cost drivers with the predefined project type.

For example, if your project is rated (according to [dwheeler cost driver table](https://dwheeler.com/sloccount/sloccount.html#cocomo)) Very High for Complexity (effort multiplier of 1.30), and Low for Language & Tools Experience (effort multiplier of 1.07), and all of the other cost drivers are rated to be Nominal (effort multiplier of 1.00), the EAF is the product of 1.30 and 1.07.

Effort Adjustment Factor = EAF = 1.30 * 1.07 = 1.391

 `scc --cocomo-project-type "organic" --eaf 1.391`

The eaf was hard-coded to a factor of 1 beforehand.

### DIFF
--- a/main.go
+++ b/main.go
@@ -114,6 +114,12 @@ func main() {
 		2.4,
 		"set the overhead multiplier for corporate overhead (facilities, equipment, accounting, etc.)",
 	)
+	flags.Float64Var(
+		&processor.EAF,
+		"eaf",
+		1.0,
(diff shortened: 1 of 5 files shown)


boyter:
Making me look bad by doing all the things I said I would do before me!

PR #300, opened 2021-11-02
Unify output of fileSummarizeShort() and fileSummarizeLong() and simple language support for number formats

fschaefer:
Basically this fixes a inconsistency in fileSummarizeShort() and fileSummarizeLong() and adds i18n support for number formats i.e. decimal and thousands separators.

```
$>./scc
...
Estimated Cost to Develop (organic) $667,319
Estimated Schedule Effort (organic) 11.79 months
Estimated People Required (organic) 5.03

$>LANG=de ./scc
...
Estimated Cost to Develop (organic) $667.319
Estimated Schedule Effort (organic) 11,79 months
Estimated People Required (organic) 5,03
```

At the moment I'm thinking about refactoring the COCOMO output to SLOCCount's output format, i.e. writing out the formula for calculating the values for better understanding of the printed values. But I'm unsure if this is something for a PR or a private branch...

### DIFF
--- a/processor/formatters.go
+++ b/processor/formatters.go
@@ -7,7 +7,6 @@ import (
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
-	"github.com/mattn/go-runewidth"
 	"io/ioutil"
 	"math"
 	"os"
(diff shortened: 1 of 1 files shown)


boyter:
Id say an option for COCOMO would be better allowing it to flip into SLOCCount's output would be better. IE opt in to it.

Looking though this now.

PR #302, opened 2021-11-04
Add --sloccount-format flag to print a more SLOCCount like COCOMO calculation

fschaefer:
As mentioned in the previous PR, I've added a flag to output a more SLOCCount like COCOMO calculation. I prefer this one because you can see the formula and values used.


processor/formatters.go
```diff
@@ -987,6 +994,23 @@ func trimNameShort(summary LanguageSummary, trimmedName string) string {
 	return trimmedName
 }
 
+func calcolateCocomoSLOCCount(sumCode int64, str *strings.Builder) {
```
cycle20, line 997: Minor typo in the function name.

PR #314, opened 2022-01-12
Add razor extension

CommonLoon102:
In Blazor the extension is `razor` instead of `cshtml` but they are completely the same thing.


boyter:
Thanks so much for the submission.

PR #316, opened 2022-02-02
Update documentation about Github action

fcsonline:
Updating the Github action tag, because there is an issue in `1.0.0` exporting the `scc` output into the Gtihub action steps.


boyter:
Nice. Merging.

PR #317, opened 2022-02-08
Add CodeQL to recognised languages

edoardopirovano:
Greetings, I hope you are doing well. I'm one of the language maintainers for [CodeQL](https://codeql.github.com/), a language for querying codebases in order to find vulnerabilities (which, in fact, you are using [here](https://github.com/boyter/scc/blob/master/.github/workflows/codeql-analysis.yml)). This PR adds CodeQL to the set of languages that `scc` will recognise and count lines for. Our language specification can be found [here](https://codeql.github.com/docs/ql-language-reference/ql-language-specification/) if you'd like to double check that what I have done is a sensible representation of that.

We have a slight vested interest in this since the CodeQL engine actually uses `scc` under the hood to do some line counting (in particular, if you go to the _Security_ tab for a GitHub repository that is using CodeQL and you are an admin of, then click _Code scanning alerts_ and the total number of lines under _Lines scanned_ is coming from `scc`). We've recently started looking into using CodeQL to analyse our own CodeQL code, so getting this PR in would allow us to get sensible line counts for our own code there.


boyter:
Neat! I did not know 

Also a very high quality PR to boot! Looks perfect to me and thank you for including a link to the language spec which makes this much easier than having to search around for it. Approved.

Should make the next version, although you can build from mainline if you need it faster.

edoardopirovano:
> Should make the next version, although you can build from mainline if you need it faster.

Thanks! We're already building from an internal fork of your repo with this patch (and also https://github.com/boyter/scc/pull/295 which one of my colleagues wrote) applied, but it's good to get it included here so when you next do a release we can update our internal fork without having to reapply any patches.

boyter:
Ah neat. I really need to finish off the last few items in the queue. I hope to have it all done for Go 1.18 release so I can get the free performance improvements that usually brings.

YASINE0:
I am a child witnessing you from Morocco. I want to learn programming from you

PR #327, opened 2022-03-24
Add Blade Template language support

nathanaelhoun:
Language documentation : https://laravel.com/docs/blade.


boyter:
Now this is something I had not anticipated. Having two languages, where they almost share the same extension `.php` but with one being `.blade.php`

I haven't checked yet, but I suspect this works randomly due to Go's map ordering and which one it hits first when processing.

Im going to have to try this out a bit to ensure it works as expected. I suspect it might need to have the keywords values populated to indicate that this is a blade file to ensure its correct. Ill add some tests to determine how likely this is first.

Thanks for the PR and including the language documentation.

PR #343, opened 2022-07-28
Add Nial, SNOBOL, and K languages

Garklein:
Notes:
- with all of K's overloads, the complexity checks will be way off. For example, `or` (`|`) is overloaded to also be `max`, so `a|b` could be read as either `a or b` or `a max b` depending on context
- both SNOBOL and Nial are completely case insensitive for keywords (ex. in SNOBOL, `output`, `OUTPUT`, and `oUTPuT` mean the same thing). I've only added the lowercase versions to the complexity check.


boyter:
Very nice! Much appreciated!

PR #363, opened 2022-12-22
Add Factor Support

Capital-EX:
This PR adds support for Factor. I made sure to include all of Factor's conditional combinators. I however did not include the generated file `constants.go` in this PR. It seems like the whole file gets regenerated, so I didn't want to cause conflicts with that. Finally, I have included an example program for SCC to be ran against.


boyter:
Thanks for the PR. Merged.

Issue #362, opened 2022-12-20
fature request: raku language support


masukomi:
I'm noting that `cloc` has it but this doesn't. I'm writing a lot of raku these days and would prefer to use this :D

boyter:
@masukomi would it be safe to copy the existing perl definitions for this? 

What extensions would we need?

masukomi:
@boyter 
> would it be safe to copy the existing perl definitions for this?

no raku's a radically different language from Perl 5. It's just by the same people ;)  Honestly i'm surprised cloc has it. 

> What extensions would we need?

If you mean "file extensions" those would be `.raku`, `.rakumod`, `.rakutest`, `.rakudoc`, `.t`

`.t` and `.rakutest` are both test files containing raku.

`.rakudoc` is, i believe, expected to contain `pod6` (an evolution of Perl 5's `pod` documentation format. Note that it's different enough that the Perl5 pod parser's don't work on it) 

note:  The old `.p6` (or `.pl6`), `.pm6` (or `.pm` which is also supported, but discouraged) and .pod6 extensions will continue to be supported for 6.e and marked as deprecated in 6.f. We're currently at `v6.d` so we've got a while before those are officially deprecated. That being said, you're not going to find them in any new codebases so I'm not sure anyone would care.  


if you mean "extensions to scc" I don't know because i don't really understand where the language parsing instructions come from. I suspect, but don't actually know, that accurately calculating complexity for Raku functions might be a major pain in the butt given some of the language features. :/

boyter:
So an example using perl (just since they are slightly related_ https://github.com/boyter/scc/blob/master/languages.json#L4981

What I need to know to add this is what extensions are in (and if duplicates, then how to keyword guess which is which) which is how it determines what the language is for processing. Im OK with adding deprecated ones since it processes older things too.

Ill see if I can find a language spec for raku when I get the chance, unless you know of one. That usually makes things much easier :)

masukomi:
after looking at that json
ahh. gotcha. I think i can put that together. Raku doesn't have a Common Lisp level spec doc (massive tome) but i think what you want can be found here: https://docs.raku.org/

The [specification page](https://www.raku.org/specification/) is really more of a central place to find the latest test suites and links to design docs.

boyter:
If you want to add it yourself let me know and you get the glory of the merged PR. Otherwise I will probably do it after christmas.

[...]

PR #364, opened 2023-01-03
added Raku info to languages.json

masukomi:
Fixes Issue #362 

## source refereneces
Source for 99% of the complexity stuff [is here in the Raku docs](https://docs.raku.org/language/control)

The all-caps ones are [Phasers](https://docs.raku.org/language/phasers) and I _think_ they should be included but I'm not 100% sure.

All the various quote marks come from [here in the grammar](https://github.com/rakudo/rakudo/blob/b5df6e28c4b1ebd2980174342c616e543c49a233/src/Perl6/Grammar.nqp#L3481)

except for `«` and `»` which are actually defined a few times in a few ways in the grammar (context alters meaning) Sometimes they function as quote marks. Sometimes they're...more. 

[edit]: added `react` and `whenever` from [the concurrency docs](https://docs.raku.org/language/concurrency#react).

## Notes 
Also of note is "emit" which is in the flow control docs but wasn't clear if it should be included for the complexity check or not.

I've also specifically not included tests like `<` and `lt` because other languages seemed to not include those. 

I've also removed perl6 from the shebang list
for perl since perl6 is a different language
(now called raku).

⚠️ there is overlap between Perl and Raku with regards to some extensions. As noted in issue #362

> The old .p6 (or .pl6), .pm6 (or .pm which is also supported, but discouraged) and .pod6 extensions will continue to be supported for 6.e and marked as deprecated in 6.f. We're currently at v6.d so we've got a while before those are officially deprecated. That being said, you're not going to find them in any new codebases so I'm not sure anyone would care.

My recommendation would be to remove those from this PR because no-one's using them anymore and a fair amount of that ancient code with old extensions doesn't work anymore anyways. Just holler and I'll make that change.

## License Declaration
I Kay Rhodes (a.k.a masukomi) explicitly license this contribution under the MIT AND Unlicense licenses.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -2157,10 +2157,10 @@
       "and",
       "or",
       "xor",
-      
+
       "eq",
       "=",
-      
+
       "smart-if",
       "smart-if*",
       "smart-when",
@@ -4499,7 +4499,7 @@
     "line_comment": [
       "%"
     ],
-    "multi_line": [ 
+    "multi_line": [
       [
         "%{",
         "}%"
@@ -4801,7 +4801,7 @@
     "line_comment": [
       "//"
     ],
-    "multi_line": [ 
+    "multi_line": [
       [
         "/*",
         "*/"
@@ -5149,8 +5149,7 @@
     ],
     "shebangs": [
       "perl",
-      "perl5",
-      "perl6"
+      "perl5"
     ]
   },
   "Plain Text": {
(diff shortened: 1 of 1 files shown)


boyter:
If you think the change to remove them, is worth it lets do that. Im happy to go along with whatever you suggest.

boyter:
Let me know if you want to do it, otherwise ill merge and all good.

masukomi:
force-pushed after adding `react {` and `whenever ` (and sorting so we could find things) based on some feedback I got. 

I'll remove those extensions tomorrow. Let's let it sit a few days to see if i can get some more eyes on it from people more experienced with Raku than me.

boyter:
Works for me. Say the word and ill merge it in.

masukomi:
Update: rebased and force-pushed with the following changes to what was there already

- removed deprecated raku file extensions
- based on [this comment](https://github.com/boyter/scc/issues/362#issuecomment-1370292864) I have added in the remaining comparators for numbers and strings
- added in the missing unicode equivalents for the various number comparators raku supports [documentation here](https://docs.raku.org/language/unicode_ascii) 
- added set comparators  

Note that I haven't added ALL of the unicode things, just the ones that match up to comparators, as things like multiplication and assignment (of which raku has many forms) don't seem to be included in `languages.json` for any language after a quick skim.

masukomi:
.... and another 

- added more multiline comment forms

Open question: Raku has something similar to Ruby's Heredocs. It's unclear how to encode that in `languages.json` so i haven't included it. 

in Raku we've got 
```raku
my $foo = q:to/END/
some stuff
END
```
and 
```raku
my $foo = qq:to/END/
more stuff
END
```

"END" could be any text. THe compiler just scans the following lines until it sees a line with only that on it. 

`q` vs `qq` at the start controls the handling of interpolation within the string. 

(there may be other `q` variants i haven't found yet.. docs are harder to find on this)

in ruby you've got  the same thing 

```ruby
query = <<-SQL
SELECT * FROM food
WHERE healthy = true
SQL
```
and 
```ruby
query = <<~SQL
SELECT * FROM food
WHERE healthy = true
SQL

`<<-` vs `<<~` is relative to how it handles indentation

boyter:
That is an edge case that currently is NOT catered for. Something to mull over. https://github.com/boyter/scc/issues/365

masukomi:
I don't expect any more feedback on this from the intarwebs and have no plans for poking it more unless requested. 

I know the JSON is valid but i'm not sure how to actually test that everything works. If you can provide me instructions on how to verify that everything is matching correctly i can put together a test file with all this stuff in it and validate it against instructions.

boyter:
If you can point me in the direction of a file, which you know the counts for I can verify myself. Otherwise adding one into https://github.com/boyter/scc/tree/master/examples/language works, as that's what I implement all the tests against.

PR #366, opened 2023-01-09
update Julia complexity checks

stevengj:
The Julia complexity checks included a `switch` statement (which [does not currently exist](https://github.com/JuliaLang/julia/issues/18285) in Julia) but omitted `elseif` and `try/catch/finally` (unlike Python). 

Also, the Julia checks oddly included two comparison operators `==` and `!=`, which are not control flow (unlike the short-circuiting `&&` and `||`), are only two of the *many* comparison operators in Julia, and don't seem to be used as complexity measures by scc for Python.  So I deleted these two from the list.   Is there a good argument for why these should be included?   (I see that a number of languages, like Ruby and C++, do include them for some reason.) 

I also considered adding Julia's `->` and `do` operators which create [anonymous functions](https://docs.julialang.org/en/v1/manual/functions/#man-anonymous-functions), but you don't use the analogous `lambda` in Python so I left it out.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -3644,13 +3644,15 @@
       "for(",
       "if ",
       "if(",
-      "switch ",
       "while ",
       "else ",
+      "elseif ",
+      "elseif(",
+      "try ",
+      "catch ",
+      "finally ",
       "|| ",
-      "&& ",
-      "!= ",
-      "== "
+      "&& "
     ],
     "extensions": [
       "jl"


boyter:
If you think they are worth adding by all means add them in. Too many languages for me to be across them all.

Thanks so much for the PR though! Will be into the next release.

stevengj:
Thanks!  (Any thoughts on why so many languages have `==` in their complexity measure?  This confused me.)

boyter:
When I added them initially, I just looked for any operation that would be a loop/branch/switch condition. Since == and != tends to result in a branch IE makes it more complex I included them.

The idea was to catch logic like

```
if input == MagicNumber && input != BadNumber 
```

Such that it would increment the count for each condition of the if since they add additional overhead to the developer.

So the following would be considered less complex

```
if input == 1
```

That was the main idea.

PR #307, opened 2021-12-13
Adding OpenMetrics report support

jan-guenter:
implements feature request #306 

Summary of code changes:
- added OpenMetrics reporter
- added unit tests for OpenMetrics reporter
- added `openmetrics` to usage help
- updated usage and added description of OpenMetrics to `README.md`
- extracted common language summary aggregation from `toCSVSummary` and `toJSON` into separate method `aggregateLanguageSummary`
- renamed `golang.org/x/text/language` import to avoid warnings about naming conflicts with local variables

PR #367, opened 2023-01-14
Make OpenMetrics output spec compliant

linusg:
The current output is valid for Prometheus, but not the derived OpenMetrics format:
- Spaces between comma-separated labels are not permitted
- The exposition must end with a `# EOF` line
- UNIT must be a suffix of the metric name, i.e. `scc_code` cannot have the unit `lines`

Additionally:
- Complexity is not measured in 'lines'

For reference, see:
https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md

I chose to remove the invalid `UNIT`s instead of renaming them to 1) not break people's monitoring and dashboards and 2) keep things consistent with the internal names, and names used in other output formats.

Ref #307, cc @jan-guenter

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,15 +627,12 @@ The output includes a metadata header containing definitions of the returned met
 # UNIT scc_lines lines
 # HELP scc_lines Number of lines.
 # TYPE scc_code count
-# UNIT scc_code lines
 # HELP scc_code Number of lines of actual code.
 # TYPE scc_comments count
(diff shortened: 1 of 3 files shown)


boyter:
Sorry lost track of this. 

Merging in, since I am presuming it to be correct :) OpenMetrics is not something I am overly familiar with.

PR #378, opened 2023-02-24
README: remove duplicated noun in "it `scc`"

ArneBab:
Now only uses `scc`.


boyter:
Thanks for fixing this.

Issue #345, opened 2022-08-12
miscount on cpp file


pagogc:
**Describe the bug**
correct counting stops at specific cpp-syntax because of wrong last state 8

**To Reproduce**
example code:
```cpp
// 1 comment

bool MyFunction(long id)
{
    MyType strBaseDir("%s\\PAN\\LAYOUTRAHMEN\\", MakeAbsolutePath("VORLAGEN"));
    // 2 comment
}
```

1. scc as command line
2. scc --trace filename.cpp
3. [test.cpp.txt](https://github.com/boyter/scc/files/9314899/test.cpp.txt)

**Expected behavior**
correct counting after line 5.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: 3.0.0 release binary

--------------------------
trace
```
TRACE 2022-08-12T08:30:44Z: milliseconds unmarshal: 1
TRACE 2022-08-12T08:30:44Z: nanoseconds build extension to language: 519400
TRACE 2022-08-12T08:30:44Z: configured to lazy load language features
TRACE 2022-08-12T08:30:44Z: nanoseconds to build language C++ features: 0
TRACE 2022-08-12T08:30:44Z: nanoseconds read into memory: test.cpp: 0
TRACE 2022-08-12T08:30:44Z: test.cpp line 1 ended with state: 1: counted as comment
TRACE 2022-08-12T08:30:44Z: test.cpp line 2 ended with state: 1: counted as blank
TRACE 2022-08-12T08:30:44Z: test.cpp line 3 ended with state: 1: counted as code
TRACE 2022-08-12T08:30:44Z: test.cpp line 4 ended with state: 1: counted as code
TRACE 2022-08-12T08:30:44Z: test.cpp line 5 ended with state: 8: counted as code
TRACE 2022-08-12T08:30:44Z: test.cpp line 6 ended with state: 8: counted as code
TRACE 2022-08-12T08:30:44Z: test.cpp line 7 ended with state: 8: counted as code
TRACE 2022-08-12T08:30:44Z: nanoseconds process: test.cpp: 1080100
-------------------------------------------------------------------------------
Language                 Files     Lines   Blanks  Comments     Code Complexity
-------------------------------------------------------------------------------
C++                          1         7        1         1        5          0
-------------------------------------------------------------------------------
Total                        1         7        1         1        5          0
-------------------------------------------------------------------------------
Estimated Cost to Develop (organic) $103
Estimated Schedule Effort (organic) 0.421026 months
[...]

PR #386, opened 2023-04-26
Fix incorrect string escaping

KAAtheWiseGit:
Previously, only one char got checked for escapes when parsing strings. This meant that strings such as "\\" never got terminated, leading to incorrect parsing.

This commit adds a check which calculates the number of escapes pre-string ending to determine wherever the last escape is itself escaped or not.

This was prompted by issue #345, I also added a test case for it.

One important note is that I am unsure how much that impacts performance.  The for loop is only accessed if the string has an escape before it, so it should be a fairly rare path.

### DIFF
--- a/examples/issue345/filename.cc
+++ b/examples/issue345/filename.cc
@@ -1,0 +1,4 @@
+bool MyFunction() {
+    func("double\\escaped\\string\\");
+    // comment
+}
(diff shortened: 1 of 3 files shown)


boyter:
Neat. Thanks for taking the time out to resolve this.

PR #389, opened 2023-05-10
Fix CI: go.yml

elliotwutingfeng:
- <https://github.com/rivo/uniseg> requires Go 1.18 for generics; use Go version found in `go.mod` for CI instead of Go 1.17.
- `actions/checkout@v2` uses Node 12 which is deprecated by GitHub; upgrade to `actions/checkout@v3`

---

This patch is licensed under MIT License and The Unlicense.


boyter:
Thanks for this. I totally forgot about dealing with that.

PR #399, opened 2023-07-28
Add support for Textile

shynur:
**Textile** is a markup language (like Markdown).

See <https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files#rendering-differences-in-prose-documents> and <https://textile-lang.com>.

Thanks in advance.


boyter:
Thanks for the submission!

PR #401, opened 2023-07-31
add support for ReScript

tsnobip:
I added support for ReScript, I basically copy-pasted from ReasonML given the syntax is very close.


boyter:
Thanks!

PR #407, opened 2023-09-12
Improve Lua and add Luau support

Heliodex:
Improves Lua complexity and comment checks, adding `--[=[ ]=]`, `--[==[ ]==]`, etc for comments as is in Lua, and multiline string support with `[[ ]]`. (unsure if this conflicts with detecting `--[[ ]]` comments)

Also adds [Luau](https://luau-lang.org/) support, essentially copy-pasted from Lua as Luau is based on it, only adding `` ` ` `` support for strings.


boyter:
Thanks so much!

PR #413, opened 2023-11-03
do not add a '~' when filename is not truncated

elindsey:
This fixes an off-by-one when the filename length exactly matches our size limit. As an example, today if the filename size limit is one character and the filename is named "a", it will be incorrectly formatted as "~a" despite no truncation occuring.


boyter:
Nice pickup. I vaguely remember thinking this could be an issue when I wrote it but was doing other stuff and the time and it slipped my mind.

Issue #143, opened 2019-12-15
Official Docker Image


Fleshgrinder:
Would it be possible that we get an official Docker image? You could host it [on GitHub](https://github.com/features/packages) like the pre-compiled binaries. Having a Docker image would simplify the usage with automation (e.g. CI).

boyter:
Yep a docker image sounds like a good idea. I will have a look to see what this involves.

Potherca:
Based on the work of [gitlab-ci-utils/docker-scc](https://gitlab.com/gitlab-ci-utils/docker-scc) (MIT License), I came up with this:

```Dockerfile
# hadolint ignore=DL3007
FROM golang:latest as scc-get

ENV GOOS=linux \
GOARCH=amd64 \
CGO_ENABLED=0

ARG VERSION=v3.0.0
RUN git clone --branch $VERSION --depth 1 https://github.com/boyter/scc
WORKDIR /go/scc
RUN go build -ldflags="-s -w"

FROM scratch
COPY --from=scc-get /go/scc/scc /scc
ENTRYPOINT ["/scc"]
```

which is basically an empty docker image with only the `scc` binary in it.

I guess the majority of the work that needs to be doen for this issue is setting up all of the right GitHub Action, correct?

boyter:
> setting up all of the right GitHub Action

That would be it as far as I can see. 

I really need to block out some time to get all of this done.

PR #425, opened 2024-01-04
Create Github action to automatically build a docker image and publish it in github package registry

lhoupert:
This should solve the issue https://github.com/boyter/scc/issues/143 and help developpers who cannot install binaries to pull directly from github a docker image of scc.

I tested the workflow here https://github.com/lhoupert/scc-in-docker/pkgs/container/scc-in-docker and it seems to be working well.

A docker image will be created each time a new push is made on main or a tag is created.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,98 @@
+name: Docker
+
+# This workflow uses actions that are not certified by GitHub.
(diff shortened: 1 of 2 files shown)


boyter:
Thanks so much for the PR. Merged.

lhoupert:
Thank you to you and your work :-)

PR #426, opened 2024-01-18
Add support for the Odin programming language

jakubtomsu:
I added support for the [Odin programming language](https://odin-lang.org/) by updating the `languages.json` file.

Tested in on my 30k loc odin codebase and the odin core library, everything seems to work fine. Not sure if comments work correctly, because Odin allows nested multiline comments unlike C/C++.

All the tests from `test-all.sh` passed, but I got a `FAILED symlink check` message (I'm assuming this is because I'm on windows using mingw64, which goofs things up somehow)


boyter:
Thanks for the PR. Merged.

PR #428, opened 2024-02-01
Fix docker image build

lhoupert:
Hi,
Sorry I realised my previous pull request was wrong as the dockerfile was not relevant for a fork. 
I did some changes and tested, it seems to be working fine on my fork: https://github.com/lhoupert/scc/pkgs/container/scc 

You can test running the v0.0.2 of docker image generated on my fork with the command below :

```
docker run --rm -it -v "$PWD:/pwd"  ghcr.io/lhoupert/scc:v0.0.2 scc /pwd
```

This should fix the github action. Each new release of scc should also create a tagged docker image

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -6,8 +6,8 @@ name: Docker
 # documentation.
 
 on:
-  schedule:
-    - cron: '36 1 * * *'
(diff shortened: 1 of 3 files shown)


boyter:
No problem. It happens. Thanks for fixing it though!

PR #430, opened 2024-02-18
remove the Gopkg.toml and Gopkg.lock files leftover from golang/dep

steverusso:
It seems these might have been overlooked when [converting to Go modules](https://github.com/boyter/scc/pull/171) about four years ago, or maybe they were left in for [`golang/dep`](https://github.com/golang/dep) users to have time to move to Go modules.

In any case, as long as I'm not missing something, `golang/dep` has been archived for over three years now, so it's probably safe to remove them?


boyter:
Weird... I thought I had removed them ages ago. Nevermind! Cheers for the PR!.

steverusso:
@boyter cheers for this awesome project! :slightly_smiling_face:

Just a quick followup question: is the `vendor` directory itself currently used at all? Or are you just keeping it around as a way to reliably have access to those external dependencies?

boyter:
It should be used by the compiler when building from source. I also keep it so I have a reliable way to access them. Also I hate projects where you clone, then spend 30 mins waiting for dependencies before being able to build.

I guess I also watched a company grind to a halt when left-pad happened while I was trucking on.

steverusso:
Haha fair enough, good points. Thank you!

PR #437, opened 2024-03-05
Add lesser-known ILPROJ extension to MSBUILD

npjg:
Many ILPROJ (Intermediate Language Project) files occur, for example, in https://github.com/dotnet/runtime. These are rather like CSPROJ files, and it seems they should be included in the MSBuild category. For some further discussion of ILPROJ, see https://github.com/dotnet/sdk/issues/16690.


boyter:
Ah cheers.

PR #442, opened 2024-03-22
Add support for JSONC (JSON with comments)

CarterLi:
Ref: https://code.visualstudio.com/docs/languages/json#_json-with-comments


boyter:
Thanks for the contribution!

PR #443, opened 2024-03-22
Add support for RC (Windows Resource-definition Script)

CarterLi:
https://learn.microsoft.com/en-us/windows/win32/menurc/about-resource-files


boyter:
Another excellent addition! Thank you.

Issue #439, opened 2024-03-11
Add support for MDX file extension


NathanielHill:
Sorry, do not have time to contribute, but wanted to add this in case it adds any value as a backlog item.

Feel free to close, will not be offended.

MDX has been around for a few years and is frequently used in frontend sites for text heavy content like blog posts.

https://mdxjs.com/docs/what-is-mdx/

boyter:
Ill have a play around later to adding this.

sgoggins:
@boyter : The syntax in these files appears to be quite similar to, if not entirely, the syntax dynamics of markdown. This might be a quick add. Here's a whole directory of files like this for quick reference: https://github.com/hashicorp/terraform/tree/main/website/docs

boyter:
Neat. Merged in. Thanks so much for adding it.

PR #445, opened 2024-04-10
Update languages.json: MDX Support

sgoggins:
I believe mdx files will be supported by adding the extension to the markdown block. The syntax *is* markdown.


sgoggins:
This fixes issue #439

boyter:
Its the `go generate` part that does it. It actually triggers this https://github.com/boyter/scc/blob/master/main.go#L14C1-L14C40 which in turn calls https://github.com/boyter/scc/blob/master/scripts/include.go which does some validation and produces embedded content.

Its common practice in Go to generate code since dynamic stuff can be a pain.

It is possible to do this via a build these days using embed however it wouldn't do the pre-checks to ensure the JSON is valid, although would be picked up later on. Mostly this is a result of being created before embed was included into Go.

I actually use the https://github.com/boyter/scc/blob/master/test-all.sh script which does it for me as well as running all tests, but this step is also included in the README https://github.com/boyter/scc/tree/master?tab=readme-ov-file#development

PR #460, opened 2024-05-17
Change FORTRAN to Fortran for modern Fortran

mathomp4:
This is a bit of a minor thing, but I was recently counting some Fortran code and saw `FORTRAN Modern` in the output. The current Standards (since Fortran 90) has used "Fortran" for the name ([cf. Wikipedia](https://en.wikipedia.org/wiki/Fortran#Naming)).

So, one of the most trivial PRs ever is made. 😄


boyter:
Thanks for fixing it. Merged.

PR #467, opened 2024-05-24
Fix Golang complexity checks

apocelipes:
Go does not have a "while" statement.
Go has "select" statements work like "switch", add this to checks.


boyter:
How in the heck did I miss that....

boyter:
Thanks so much for fixing it.

PR #473, opened 2024-05-30
Update README.md

tkothe:
updated link to Java complexity conditions JSON


boyter:
Thanks. One of those ones I think I need to totally change to work as I don't want to update it every single time.

PR #474, opened 2024-06-02
Add `pnpm-lock.yaml` to default exclude file list

gtn1024:
`pnpm-lock.yaml` is the file just like `package-lock.json` or `yarn.lock`


boyter:
Neat cheers for that.

PR #482, opened 2024-06-16
improve: remove redundant calls of sprintf

apocelipes:
- replace `fmt.Println(fmt.Sprintf(...))` with `fmt.Printf`, this is simpler.
- replace `sb.WriteString(fmt.Sprintf(...))` with `fmt.Fprintf`, this is simpler and much faster:
```text
$ benchstat old.bench new.bench
goos: windows
goarch: amd64
pkg: testbuilder
cpu: Intel(R) Core(TM) i5-10200H CPU @ 2.40GHz
          │  old.bench  │              new.bench              │
          │   sec/op    │   sec/op     vs base                │
Builder-8   1.808µ ± 2%   1.533µ ± 1%  -15.19% (p=0.000 n=10)

          │  old.bench   │              new.bench               │
          │     B/op     │     B/op      vs base                │
Builder-8   1.281Ki ± 0%   1.000Ki ± 0%  -21.95% (p=0.000 n=10)

          │  old.bench  │             new.bench              │
          │  allocs/op  │ allocs/op   vs base                │
Builder-8   15.000 ± 0%   6.000 ± 0%  -60.00% (p=0.000 n=10)
```

Benchmark code is here:
```go
// the old one
func BenchmarkBuilder(b *testing.B) {
	for range b.N {
		var sb strings.Builder
		for i := range 10 {
			sb.WriteString(fmt.Sprintf("%d. writing an integer number: %d", i, i))
		}
	}
}

// the new one
func BenchmarkBuilder(b *testing.B) {
	for range b.N {
		sb := &strings.Builder{}
		for i := range 10 {
			fmt.Fprintf(sb, "%d. writing an integer number: %d", i, i)
		}
	}
}
```

### DIFF
--- a/processor/formatters.go
+++ b/processor/formatters.go
@@ -492,31 +492,31 @@ func toOpenMetricsSummary(input chan *FileJob) string {
 	language := aggregateLanguageSummary(input)
 	language = sortLanguageSummary(language)
 
-	var sb strings.Builder
+	sb := &strings.Builder{}
 	sb.WriteString(openMetricsMetadata)
 	for _, result := range language {
-		sb.WriteString(fmt.Sprintf(openMetricsSummaryRecordFormat, "files", result.Name, result.Count))
-		sb.WriteString(fmt.Sprintf(openMetricsSummaryRecordFormat, "lines", result.Name, result.Lines))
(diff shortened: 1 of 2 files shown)


ccoVeille (approved):
LGTM 👍 thanks

PR #497, opened 2024-08-24
chore: fix github-action warnings and errors

apocelipes:
To fix warnings and errors like these:
image
image

Also fixed some Dockerfile warnings.


boyter:
Thanks for that.

PR #504, opened 2024-09-05
openmetrics formatter: emit gauge metrics

jan--f:
`count` was not a valid metric type and `gauge` is the better type since counters are expected to strictly increase.

An up to date Prometheus will complain about unknown metric types. 
See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#type for valid types.

The doc says:
> Gauges are current measurements, such as bytes of memory currently used or the number of items in a queue. For gauges the absolute value is what is of interest to a user.


boyter (approved):

boyter:
Thanks for that. Its something I am woefully ignorant about.

PR #505, opened 2024-09-06
fix: fix OpenMetrics tests

apocelipes:
#504 changed the OpenMetrics outputs but forgot changing the tests. Fit these tests into the correct output formats.

image


boyter:
My bad. I should have checked it, but was in a hurry. Thanks for picking it up.

PR #506, opened 2024-09-10
feat: enhance awk,c,cpp complexity checks

apocelipes:
- add AWK complexity checks
- enhance c/cpp complexity checks, support these forms:
```c++
while(1) {
//
}

switch(num) {
case 1:
   ...
default:
   ...
}
```
- clean up test-all.bash, put specific languages into an array that can be easily maintained


boyter (approved):

boyter:
Thanks for the submission. Redoing the test-all.sh is one of those things thats on my list. I want to move it all to pure Go code in the future just to make things easier to maintain.

PR #509, opened 2024-09-11
feat: support --exclude-content

### LINKED ISSUE #507: Feature request:
Exclude autogenerated files based on an included comment/string.

cloc allows --exclude-content="Code generated by sqlc. DO NOT EDIT."

Something similar would be great

apocelipes:
Fixes #507.

Support the "--exclude-content" option.

The reason why I put file content filtering in scc instead of gocodewalker is that we need to read the file contents first, and scc already did it, so only a small amount of code needs to be modified to implement the "exclude-content". And I put the filter in `fileProcessorWorker()` because that makes it run in different goroutines in parallel.

**When we set "--exclude-content", the performance could get about 20% downgrade because we need to check files contents line by line:**

benchmark

_There's no performance impact when not adding the "--exclude-content" option._

### DIFF
--- a/SCC-OUTPUT-REPORT.html
+++ b/SCC-OUTPUT-REPORT.html
@@ -12,14 +12,14 @@
 	</tr></thead>
 	<tbody><tr>
 		<th>Go</th>
-		<th>30</th>
-		<th>9515</th>
-		<th>1460</th>
-		<th>456</th>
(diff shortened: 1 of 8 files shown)


apocelipes:
This PR is no more needed since 507 has been resolved.

PR #515, opened 2024-09-27
Add example of creating public URL

danshearer:
This is something I do fairly often, so I thought I'd share


boyter:
Thanks for taking the time to contribute it back.

PR #516, opened 2024-09-27
Enhance scripts/include.go and fix warnings

apocelipes:
Write tmp data to a buffer rather than a temp file. This is faster and robuster.

Fix some golangci-lint warnings.

Fix a miss use of break, a break in a switch only stop the switch statment, to stop the outer for-loop we need a label.


boyter:
Thanks for this.

This was something I was starting to look at actually as id like to remove the need for JSON at all inside the application. This is a good stepping stone to it.

PR #533, opened 2024-10-01
Tell GitHub that `.sccignore` is an ignore list

spenserblack:
This is a super-nitpicky PR, because I like to see syntax highlighting when I look at a file :sweat_smile: I think you're already familiar with linguist, but just to describe what this PR does:

- Applies Ignore List syntax highlighting to `.sccignore`
- If for whatever reason you decided to include Ignore List stats in the GitHub language bar, `.sccignore` would be counted to that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.sccignore linguist-language=Ignore-List


boyter (approved):

boyter:
Thanks for the PR. Merged. I should look into how linguist does this more as something similar for scc might make it more useful for integrations.

spenserblack:
Regarding how Linguist does it, I don't think it does much more than a case insensitive match to one of its language's names or its aliases. It does parse out the `.gitattributes` (using git2 IIRC) but, for SCC, IDK if `.gitattributes` should be the "official" place to set language overrides. Linguist is very special in that it's used for *Git*Hub and analyzes *Git revisions,* not files (e.g. you can delete all your files from your repo and still use it). So it makes a lot of sense to tightly couple with Git tools. Since SCC is more generic, and isn't Git-specific, perhaps it can pioneer a more generic `.attributes` file (or `.sccattributes`)? But you also have the freedom to do something completely different.

I'm going on a bit of a rant, but I just want to make the point that, while Linguist is certainly a very cool tool, what's best for Linguist might not be best for SCC.

spenserblack:
Oh, just an FYI that the generic `.ignore` will be handled by https://github.com/github-linguist/linguist/pull/7070, so you don't need to worry about that.

PR #537, opened 2024-10-07
Make winget install more specific

spenserblack:
I'm basically just copying good instructions that I've seen in other projects. What this does is add:

- `--id` option that ensures that `benboyter.scc` is the ID. Without this `benboyter.scc` is getting passed as a query instead of as a unique ID. This could potentially cause a similarly named package to get installed.
- `--source` option to specify that the `winget` repository should be used.

Personally, I *always* use the `--id` option with `winget`.


boyter:
Neat. Thanks for this.

PR #542, opened 2024-10-21
refactor: using standard libraries to do file operations

apocelipes:
Using standard libraries to get temp path and to do file operations.

This makes the program more simple and safer because standard libraries clean the path and no need to use outside commands.

It also makes the code more portable since not all systems use '/tmp' as its tempfile path (other places for example, '/data/local/tmp'). Many mordern Linux distros also support setting the env varibale '$TMPDIR' to change a program's tempfile path. So using the standard library's getTempDir function is a better idea, it's done all the works for us.

### DIFF
--- a/SCC-OUTPUT-REPORT.html
+++ b/SCC-OUTPUT-REPORT.html
@@ -13,12 +13,12 @@
 	<tbody><tr>
 		<th>Go</th>
 		<th>27</th>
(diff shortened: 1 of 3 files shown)


boyter:
Merging in although I never plan on running the Python script again and the Go code is exclusively running on linux. Thanks for the PR though!

PR #543, opened 2024-10-22
Support `licen(c|s)e-apache`

spenserblack:
It's a pretty common filename: https://github.com/search?q=path%3A%2F%28%5E%7C%5C%2F%29LICEN%28S%7CC%29E-APACHE%24%2F&type=code

You'll often see it in Rust projects, which are often MIT/Apache-2.0 dual-licensed.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -3504,6 +3504,8 @@
       "copying3",
       "unlicense",
       "unlicence",
+      "license-apache",
+      "licence-apache",
       "license-mit",
       "licence-mit",
       "copyright"


boyter:
Thanks for the submission. Will make the next release.

Actually license rules are something I want to revisit at some point, based on the logic used in https://github.com/boyter/lc but turning those rules into something that works using JSON.

spenserblack:
Interesting! I'm sure you've already thought about this, but in another project I've worked on there were issues detecting if the license was "GPL X.Y Only" or "GPL X.Y or Later" (https://github.com/o2sh/onefetch/issues/1251). Just a heads up for when you're planning implementation.

boyter:
For scc its only attempting to identify there is a license file, not the contents of it. Finding the version is a whole other issue I need to get back to working on again one of these days.

PR #545, opened 2024-10-24
Mark processor/constants.go as an automatically generated file

apocelipes:
Add the 'DO NOT EDIT' mark to it, this can help developers. This is also one of the common practices in the go community.

Re-generate files because #543 that changed the language.json didn't update them.


boyter:
One of those things I never thought about. Thanks for this.

PR #548, opened 2024-10-27
Simplify and clean up the code

apocelipes:
- simplify the usage of switch statements
- use "lang" and "langs" as aliases for "language" (these aliases are already used in existing code)
- reuse sortLanguageSummary
- use gettempdir to fix a warning about directly using the '/tmp' path


boyter:
Neat. Looks good. Thanks for another one.

Issue #550, opened 2024-10-29
failing to use `.sccignore` 


janpfeifer:
**Describe the bug**
Follow up of #549 , but the issue now is that `.sccignore` is not processed.

Example: my project has all the `.proto` and generated code in a sub-directory `internal/protos`:

```bash
$ cat .sccignore
internal/protos
$ scc
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
Go                          15      5339      553       827     3959        632
Protocol Buffers             3      1186      190       691      305          0
Jupyter                      2       512        0         0      512          0
Markdown                     2        20        7         0       13          0
License                      1       201       32         0      169          0
gitignore                    1        29        7        11       11          0
───────────────────────────────────────────────────────────────────────────────
Total                       24      7287      789      1529     4969        632
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop (organic) $145,437
Estimated Schedule Effort (organic) 6.610459 months
Estimated People Required (organic) 1.954619
───────────────────────────────────────────────────────────────────────────────
Processed 288291 bytes, 0.288 megabytes (SI)
───────────────────────────────────────────────────────────────────────────────

$ scc --by-file | grep internal/protos
~internal/protos/onnx-ml.pb.go      3208      345       522     2341        316
internal/protos/protos.go              4        1         2        1          0
~internal/protos/onnx-ml.proto       895      132       524      239          0
internal/protos/README.md             11        4         0        7          0
```

**Expected behavior**
The same it does with `.ignore` which is currently working:

```bash
$ mv .sccignore .ignore
$ cat .ignore 
internal/protos

$ scc
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
Go                          11      1139      100       139      900        228
Jupyter                      2       512        0         0      512          0
License                      1       201       32         0      169          0
Markdown                     1         9        3         0        6          0
gitignore                    1        29        7        11       11          0
ignore                       1         1        0         0        1          0
───────────────────────────────────────────────────────────────────────────────
Total                       17      1891      142       150     1599        228
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop (organic) $44,221
Estimated Schedule Effort (organic) 4.204901 months
Estimated People Required (organic) 0.934321
[...]

Issue #532, opened 2024-10-01
go install github.com/boyter/scc/v3@latest isn't working since v3.4


PhilKrau:
Hey, I'm using scc inside a gitlab ci/cd-pipeline and all my pipelines are failing since the command ```go install github.com/boyter/scc/v3@latest``` isn't working anymore. Up until version 3.3.5 everything ran fine, now I get the following errors:
``` 
/go/pkg/mod/github.com/boyter/scc/v3@v3.4.0/processor/detector.go:6:2: package cmp is not in GOROOT (/opt/bitnami/go/src/cmp)
/go/pkg/mod/github.com/boyter/scc/v3@v3.4.0/processor/bloom.go:3:8: package math/rand/v2 is not in GOROOT (/opt/bitnami/go/src/math/rand/v2)
/go/pkg/mod/github.com/boyter/scc/v3@v3.4.0/processor/detector.go:9:2: package slices is not in GOROOT (/opt/bitnami/go/src/slices)
```
I would appreciate any kind of help since I'm not too versed in Go.
I'm using Ubuntu 22.04 and go1.18.1 linux/amd64.
Thanks in advance for any help

boyter:
Version 3.3.5 was updated to use Go 1.22 since the version it was on at the time was due to be EOL. 

All you would need do is update your Go version 1.22 or later (1.23 being the latest at the moment). 

I will be updating it every now and then simply to ensure it gets the latest security patches, features and performance improvements. For the future checking https://github.com/boyter/scc/blob/master/go.mod will let you know what version you need to be building from.

PhilKrau:
Managed to fix the issue, thank you very much !

boyter:
You are welcome!

PR #551, opened 2024-10-29
Update the install section of the README.md

apocelipes:
Updates #550 and #532.


boyter:
Again, thanks for keeping on top of things while I am sleeping.

PR #553, opened 2024-11-03
languages: Add SPL

TudbuT:
This adds support for the SPL language.


boyter:
Thanks so much for the submission. Will make the next release.

PR #556, opened 2024-11-06
Use json-iterator/go for all JSON operations

apocelipes:
Currently, we have two JSON libraries in the code base: "encoding/json" and "github.com/json-iterator/go". Use one of them is completely enough.

The "json-iterator/go" is faster than the stdlib and uses less memory. So we choose the fast one for all JSON operations.

It also reduces the binary file size. For an example on Linux AMD64 with Go1.22.7, the binary file change is `8988568 -> 8811060 bytes`, more than 150 kib was reduced.

### DIFF
--- a/SCC-OUTPUT-REPORT.html
+++ b/SCC-OUTPUT-REPORT.html
@@ -13,13 +13,13 @@
 	<tbody><tr>
 		<th>Go</th>
 		<th>27</th>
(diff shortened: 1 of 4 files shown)


boyter:
Reduction of the binary size is the main reason I like this. The reason I removed it was due to the speed being negligible for everything I tried.

PR #567, opened 2024-12-19
Update Vala's language information

apocelipes:
Support verbatim strings, formatted string literals and the "foreach" keyword.


boyter:
Thanks. Will make the next release sometime next year.

PR #571, opened 2025-01-14
Add Treetop Support - https://cjheath.github.io/treetop/

eric1234:
A DSL that integrates with Ruby for expressing a PEG (parsing expression grammer). The tool uses itself to define the language so if you really want the full details on the language it is at:

https://github.com/cjheath/treetop/blob/master/lib/treetop/compiler/metagrammar.treetop

But the cliff notes are:

* Very Ruby-like
* Only supports single-line comments
* Embeds proper Ruby code so I just copy/pasted the complexitychecks from the Ruby language specs.

To make this commit I just:

1. Added it to the `languages.json` file (per README)
2. Ran `go generate` (per README)
3. Ran `./test-all.sh`. This confirmed all green and updated `LANGAUGES.md` and `SCC-OUTPUT-REPORT.html` (maybe this step should be added to README?)

I also ran `go build` to create the binary and tested it against my project which has many Treetop files and confirmed it worked.


boyter:
One of the best PR's I have ever seen. Not only approved but noted as one for others to look at. Thank you very much!

Will make the next release.

PR #572, opened 2025-01-14
Make "--exclude-dir" support multi-level dirs

### LINKED ISSUE #564: exclude-dir doesn't work with multi-level dir
**Describe the bug**
exclude-dir does not work with multiple-level paths, such as: scc --exclude-dir level1\level2

**To Reproduce**

1. Tell us how you were using `scc` either as command line or library.
command line

2. If command line please list the arguments. If library please supply code to demonstrate the issue.
scc --exclude-dir level1\level2
(also tried scc --exclude-dir level1/level2 - back slash vs forward slash)

3. If the issue is a miscount or some such please link or supply a file which demonstrates the issue.

**Expected behavior**
Given a structure such as:
\MyRepo
..program.cs
....\level1
......\level2
........MyReport.rdl

"scc --exclude-dir level1\level2" should not include MyReport.rdl (i.e. \MyRepo\level1\level2\MyReport.rdl).

 - Windows
 - 3.4.0
 
** Additional questions for completeness
- Can / or \ be used as the path separator? ..or is it dependent on OS?
- Can I include quotes around the exclude-dir value when there are spaces in the paths, such as:
  "scc --exclude-dir "level1\level2,level1\Stored Procs"?

apocelipes:
Fix #564.

Upgrade modules.

Test case:
image


boyter:
Thanks for doing that @apocelipes as always. Really helpful as I am a little busy with work right now.

PR #577, opened 2025-01-25
Elixir HEEx template support, Elixir complexity

serpent213:


boyter:
Thanks for this. I reverted the change to constants.go which is updated by go build for this, to get past the conflicts, but will ensure it is updated fairly soon, and certainly before the next release.

This should make the next version when the new version of Go is released, otherwise you can build from source if you want it now.

PR #578, opened 2025-01-31
Add C++ module file extensions

MarkusBauer:
With the adaption of [C++ modules](https://en.cppreference.com/w/cpp/language/modules), more file extensions for C++ code appeared. 
This PR adds the most common ones (i.e., the list supported by CLion).

I explicitly license my contribution under MIT license.


boyter:
Thanks for this. I undid the result of the go build in this case due to conflicts but that will be rerun before a release is pushed out and will also be done and pushed once I get all the other pending pr's resolved.

Will make the next release when the next version of Go is released, otherwise you can build from source to have it now.

PR #588, opened 2025-02-12
Upgrade to Go 1.24

apocelipes:
I think it's time to upgrade to Go v1.24. Using the new API `strings.SplitSeq` and the new implemention of builtin maps, we can gain about 5% performance improvements:

benchmarks

The 1.24 release note also says that `sync.Map` should have a better performance, but I have no enough time to do benchmarks, so this PR does not touch that part of the code.

Dependencies were also upgraded.

Edit: This PR should wait for the new docker images.


boyter:
I have been playing around with 1.24 and it does look like the changes produce some repeatable albeit small gains. Once the docker build works we can look at adding this in. The moment it works I will cut a new release too.

apocelipes:
Well, "golang:1.24.0-alpine3.20" can be pulled now but "mcr.microsoft.com/devcontainers/go:1.24" is still not available. That means https://vscode.dev/ can not be used on this repo for a few days until they update the image.

So shall we wait for "mcr.microsoft.com/devcontainers/go:1.24" or just upgrade the go version?

boyter:
Oh very nice. Lets make it real.

apocelipes:
All the tests passed, so I guess I can remove the DRAFT status. `.devcontainer/devcontainer.json` can be updated in another pr when the MS updates their images (it usually takes days).

Issue #337, opened 2022-05-17
[feature] shell completion


Freed-Wu:
Now, scc has many command options. If it can provide shell completion (bash/zsh/fish/...), it will be more convenient to use. Thanks.

boyter:
@Freed-Wu Do you have any details on how to achieve such a thing?

Freed-Wu:
Maybe <https://docs.rs/clap_complete/latest/clap_complete> can be helpful?

Or that projects support completions (e.g. ripgrep, fd, exa, github-cli, etc) can be a reference.

boyter:
Ah... thats what you mean. I think this is pretty easy https://jmtirado.net/shell-completion-with-cobra-and-go/

lihaohong6:
I gave this a try in #606. I'm not familiar with go, and the implementation is pretty hacky, but hopefully a solution will come out of it.

boyter:
Thank you @lihaohong6 

@Freed-Wu if you are able to verify this, that would help out a lot.

boyter:
TODO to finish this off is update the README.md with a brief guide. I shall do that once I play around with it a bit more.

lihaohong6:
Once a new version with shell completions is released, it'll also be helpful to notify downstream maintainers so that they can install completions as a part of the package. I'll use the following for Fedora, and I would assume that packagers for other package managers can do something similar. 
```
SCC=./bin/scc
$SCC completion --shell bash > scc.bash
$SCC completion --shell fish > scc.fish
$SCC completion --shell zsh > _scc
install -Dpm 0644 scc.bash -t %{buildroot}%{bash_completions_dir}
install -Dpm 0644 scc.fish -t %{buildroot}%{fish_completions_dir}
install -Dpm 0644 _scc     -t %{buildroot}%{zsh_completions_dir}
```

boyter:
I think this is done... I dont maintain the downstream consumers, but hopefully they notice.

PR #606, opened 2025-04-15
Print command line completions

lihaohong6:
This is a prototype for command line completions (#337). 

Doing it the same way as other applications potentially interferes with existing `scc` command line argument parsing. In particular, users may want to count the LOC in the `completion` sub-directory, so `scc completion` would lead to unexpected results. 

I made it so that command line completions are generated only if the arguments are `scc completion --shell [name of shell]`. The `--shell` option should prevent users from accidentally triggering this feature. It's nonetheless still a pretty ugly implementation. Perhaps a more elegant one exists?

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"runtime"
@@ -11,6 +12,21 @@ import (
 	"github.com/boyter/scc/v3/processor"
 	"github.com/spf13/cobra"
 )
+
+func printShellCompletion(cmd *cobra.Command, command string) error {
+	switch command {
+	case "bash":
+		return cmd.GenBashCompletionV2(os.Stdout, true)
+	case "zsh":
+		return cmd.GenZshCompletion(os.Stdout)
+	case "fish":
+		return cmd.GenFishCompletion(os.Stdout, true)
+	case "powershell":
+		return cmd.GenPowerShellCompletion(os.Stdout)
+	default:
+		return errors.New("Unknown shell: " + command)
+	}
+}
 
 //go:generate go run scripts/include.go
 func main() {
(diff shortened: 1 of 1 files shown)


boyter:
OK I have no idea how this works and thus no way to verify its doing what it needs.

How do I trigger this to verify it?

lihaohong6:
It depends on the shell (and possibly the operating system). I'm using Linux through WSL to experiment, so hopefully the procedure is more-or-less the same on macOS if that's what you use. 

For `zsh`, which is what I usually use, I would

1. Run `./scc completion --shell zsh > _scc` to generate shell completions and pipe the output to a file. Underscore + name of executable is usually how `zsh` completion files are named.
2. Run `sudo cp _scc /usr/share/zsh/site-functions/_scc`. This copies the completions file to the location where `zsh` completion scripts are stored. There are ways to have `zsh` load completion scripts without superuser privileges, but this is the easiest. 
3. Restart the shell.
4. Type `scc --no-` and then press tab. Something along the lines of the screenshot below should appear.
image

I have a bunch of `zsh` plugins, so hopefully this applies to a vanilla `zsh` experience too.

Other shells are pretty similar. For `bash`, the command would be `./scc completion --shell bash > scc.bash`, and the file would be stored in `/usr/share/bash-completion/completions`. `bash` requires a double tab to autocomplete, and shows something similar. I haven't customized `bash`, so this should be easier to reproduce. 

image

For `fish`, the command is `./scc completion --shell fish > scc.fish` and the file location is `/usr/share/fish/vendor_completions.d`. My `fish` is a fresh install, so it should be easily reproducible.

image

Technically `powershell` is supported too, but I'm not familiar with that ecosystem. Since everything is autogenerated anyway, I would assume `cobra` did it correctly.

boyter:
Ah gotcha. Im going to merge then and mess around with it. Thanks for this. We can keep track on the linked issue.

PR #609, opened 2025-04-22
Support xmake & update lua and luau

apocelipes:
Support XMake. XMake is a Lua based DSL for the xmake build system.

XMake homepage: <https://xmake.io/#/>

Clean up and update Lua/Luau language informations.


boyter:
As always thank you.

Issue #611, opened 2025-05-03
Line comments are counted as code for Nix


AndrezHP:
**Describe the bug**
Line comments in Nix are not counted as comments since it uses # and not //.
Instead they are counted as code.

**To Reproduce**
Have any nix file:
file.nix
{
  \# comment
}

Call: scc file.nix
-> Comments = 0
-> Code = 3

**Expected behavior**
Commented lines are counted as comments.

**Desktop (please complete the following information):**
 - OS: [Linux]
 - Version [NixOS 25.05]

boyter:
Merged. Thanks for fixing it.

PR #612, opened 2025-05-03
Fix line comment count for nix (#611)

AndrezHP:
Fixes the issue of line comments not being counted by using \# as the line comment symbol instead of //.
https://github.com/boyter/scc/issues/611

This contribution is licensed under the present MIT License.


boyter:
Thanks for that!

PR #616, opened 2025-05-13
Support GTK Blueprint

apocelipes:
[Blueprint](https://gnome.pages.gitlab.gnome.org/blueprint-compiler/reference/index.html) is a markup language for GTK user interface files.


boyter:
Neat. Always nice to see more languages. Cheers.

PR #617, opened 2025-05-20
Support new language Go+

apocelipes:
[Go+](https://goplus.org/) is a dialect of golang and is designed for engineering, STEM education, and data science.


boyter:
As always thanks for the contributions.

PR #621, opened 2025-06-16
feat: Support .Net 10 CSharp shebang

apocelipes:
.Net 10 allows running a single C# file without project config. Here is the example: <https://devblogs.microsoft.com/dotnet/announcing-dotnet-run-app/#using-shebang-lines-for-shell-scripts>


boyter:
Neat. Im still reading though that epic post about performance improvements in 9.0. 

Thanks for this.

PR #622, opened 2025-06-19
feat: format numbers using current locale

### LINKED ISSUE #335: Option for human-readable format
First, thanks for this great tool!

For large code bases, the numerical output are hard to read, example:
image

It would be nice to get an option for human readable format, like 72.9 M (or 72.9e6)

gselig1a:
Makes it more readable for short summary, just like for the "Estimated Cost to Develop" like:
10,000,000 instead of 10000000
fixes #335

## Before
image

## After
image

+ Markdown fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,12 @@
-Sloc Cloc and Code (scc)
-------------------------
+# Sloc Cloc and Code (scc)
 
-<img alt="scc" src=https://github.com/boyter/scc/raw/master/scc.jpg>
+[![SCC illustration](./scc.jpg)]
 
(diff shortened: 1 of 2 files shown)


README.md
gselig1a, line 1: Lot of small fixes to make Markdownlint happy.

boyter: One of those things I have been kicking down the road forever. Thanks for actually fixing it.

boyter:
Thanks for that @gselig1a ill ensure it makes the next release which should come right after the next version of Go is released.

PR #624, opened 2025-06-22
feat: speed up specific languages test

apocelipes:
Run `./scc "examples/language/" --no-scc-ignore` once then reuse its result on the specific languages test. Otherwise we would need to run SCC hundreds of times over and over in an increasingly complex directory, and even though SCC is very fast it would significantly increase the time spent on testing.

The time taken to run `test-all.sh` on my laptop went down from 11.5s to 9.7s.


boyter:
Thanks. I actually have on my TODO list at some point to rewrite the while test portion to just more Go to make these sort of things a bit easier.

PR #626, opened 2025-06-23
fix: fix CountStats performance

apocelipes:
The compiler is still not smart, branch-less code in `CountStats` casused a performance regression. I'll revert this part.

Other parts won't impact the speed.

Before:
slow

Fixes:
fix

After the revert the performance issue is fixed and we even become a little bit faster.

The compiler reports that these functions are inlined (I also detected the binary file, printXXX functions actually inlined and they were not in the symbol table), and theoretically there should be no performance difference. This is so wierd, so I left some comments on it.


boyter:
Yeah annoying.

PR #628, opened 2025-06-28
feat: reduce allocs in DetectSheBang

apocelipes:
Reduces the memory allocation in the function `DetectSheBang`.

Also make git ignore the ".vscode" directory so that some private editor settings don't get added to the repository.


boyter:
Cheers.

PR #648, opened 2025-09-29
perf: speed up toCSVSummary

apocelipes:
Optimizations:
- Sorting summaries is faster than sorting string arrays, so sorting summaries before serializing.
- Reusing memory reduces the performance overhead when serializing to CSV.

These optimizations can gain performance improvements:

```
goos: darwin
goarch: arm64
pkg: github.com/boyter/scc/v3/processor
cpu: Apple M4
                 │     old     │                 new                 │
                 │   sec/op    │   sec/op     vs base                │
FileSummerize-10   1.696µ ± 1%   1.467µ ± 6%  -13.51% (p=0.000 n=10)

                 │     old      │                 new                  │
                 │     B/op     │     B/op      vs base                │
FileSummerize-10   5.375Ki ± 0%   4.828Ki ± 0%  -10.17% (p=0.000 n=10)

                 │     old     │                new                 │
                 │  allocs/op  │ allocs/op   vs base                │
FileSummerize-10   12.000 ± 0%   7.000 ± 0%  -41.67% (p=0.000 n=10)
```

This is the case where only two languages are involved. The more languages included in the result, the larger the improvement achieved.

### DIFF
--- a/SCC-OUTPUT-REPORT.html
+++ b/SCC-OUTPUT-REPORT.html
@@ -13,13 +13,13 @@
 	<tbody><tr>
 		<th>Go</th>
 		<th>27</th>
-		<th>23110</th>
-		<th>1442</th>
-		<th>442</th>
-		<th>21226</th>
-		<th>1369</th>
-		<th>449731</th>
-		<th>6246</th>
+		<th>23053</th>
+		<th>1444</th>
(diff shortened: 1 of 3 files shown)


boyter:
Thank you.

PR #649, opened 2025-10-04
fix: Remove ComplexityLine from JSON output

### LINKED ISSUE #645: Remove ComplexityLine from JSON output
Should not be there, as its not counting it so its just bloating the JSON response.

Needs to be done before next release.

To replicate

```
scc --by-file -f json
```

apocelipes:
Fixes #645

Reduce `Language` size from 208 to 200 bytes by reordering fields. This can save some memory since we already have 350+ languages.


boyter:
Cheers. Been slammed recently and had not had the chance to do this despite it being so simple.

I think a few more checks and new release will be cut.

Issue #582, opened 2025-02-06
Feature Request: Badge Color Words to simplify


HariSekhon:
On shields.io you can specify colours using a simple colour name instead of hex code:

For example:

```
https://img.shields.io/badge/Hari-Sekhon-blue?logo=github&logoColor=white

```

Becomes a blue background on the label and a white background on the logo:

![](https://img.shields.io/badge/Hari-Sekhon-blue?logo=github&logoColor=white)

This looks nicer in terms of both readability and simplicity of not having to look up hex codes and also the shields simple colours are really nice too and worth copying IMO.

`blue` is simpler than `#2081C2`

So this ticket is a feature request to add that simplified colour usage for the color parameters for scc badges.

PR #669, opened 2026-01-06
Issue #582: Add simplified badge colors

demarsdouglas:
Hello! Longtime fan of this project and excited to contribute! :)

This PR looks to address this issue: https://github.com/boyter/scc/issues/582

I've added support for named colors in `scc` badges. This lets end users use human readable color names instead of only hex values, and preserves backwards compatibility.

**WHAT CHANGED**:
- Added a `namedColors` map that covers the following:
    - shields.io colors (brightgreen, green, yellowgreen, etc..)
    - semantic aliases (success, important, critical, etc..)
    - standard CSS color names (100+ common web colors like white, black, navy, teal, etc..)
- Added a `resolveColor()` helper
    - Accepts either a named color or a hex value
    - Returns a normalized hex value
    - Is case-insensitive
    - Returns an empty string for invalid inputs
- Updated `parseBadgeSettings()` to use `resolveColor()` instead of the old hex-only regex validation
- Added tests :)

With the above, users can now write:
```
?badge-bg-color=blue&title-bg-color=brightgreen&font-color=white
```
instead of
```
?badge-bg-color=007ec6&title-bg-color=44cc11&font-color=fff
```

Hex codes work exactly as before so this is fully backwards compatible.

Let me know if you have any feedback and I can implement it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1084,19 +1084,37 @@ For `cocomo` you can also set the `avg-wage` value similar to `scc` itself. For 
 
 Note that the avg-wage value must be a positive integer otherwise it will revert back to the default value of 56286.
 
-You can also configure the look and feel of the bad using the following parameters,
+You can also configure the look and feel of the badge using the following parameters,
 
 - ?lower=true will lower the title text, so "Total lines" would be "total lines"
-he below can control the colours of shadows, fonts and badges
+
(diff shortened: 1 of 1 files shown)


boyter:
Done. Merged and deployed. 

Thank you so much @demarsdouglas for this! It was one of those things I have been meaning to add in forever.

PR #671, opened 2026-01-12
Add Apex (Salesforce) support

fffej:
Apex is like Java, except it uses single quotes for strings (I'm sure there's lots of other differences too!).


boyter:
Thank you so much for adding it!

PR #674, opened 2026-02-01
feat: support new typescript extensions: cts, mts

yifanwww:
4 years ago TypeScript introduced new file extensions `.cts` and `.mts` on version 4.7, this PR is to add the support

For details of the relationship between `.cts`, `.mts` and `.ts`, please refer to [module-format-detection].

[module-format-detection]: https://www.typescriptlang.org/docs/handbook/modules/reference.html#module-format-detection


boyter:
Thanks for the submission!

PR #687, opened 2026-03-23
Reduce string allocations

qwerty8811:
**Summary**
Reduce avoidable allocations in language detection and remap matching by switching substring checks from string-based matching to byte-based matching.

- The previous implementation repeatedly converted `[]byte` content to `string` just to perform substring checks.
- This added unnecessary allocations

**Changes**
- In detector.go, avoid converting file content to `string` during ambiguous language detection.
- In processor.go and structs.go, precompute per-language keyword byte slices once during language feature setup.
- In workers.go, switch `RemapAll` and `RemapUnknown` content checks from `strings.Contains(string(...))` to `bytes.Contains(...)`, avoiding copies of the file-content slice on each check.
- Keep behavior unchanged while reducing repeated content-copy allocations.

### DIFF
--- a/processor/workers_test.go
+++ b/processor/workers_test.go
@@ -1948,3 +1948,86 @@ func BenchmarkCountStatsWithClassify(b *testing.B) {
 		CountStats(&fileJob)
 	}
 }
+
+func BenchmarkHardRemapLanguage(b *testing.B) {
+	job := &FileJob{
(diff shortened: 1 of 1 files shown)


boyter:
Oh very nice. Looking though this in detail now. It might need some benchmarking for me first just to ensure its all good, but it looks promising.

Issue #396, opened 2023-07-09
Recognize file type base on mime type


nkh:
I have a bunch of bash files that are not counted in because they don't have a shebang

using the file type or the mime type would make them part of the scc

boyter:
Could you provide an example of a file (2 would be better) that show how you would expect this to work. I want to see the file itself to determine how this should work.

nkh:
Here's a file, let me know if you need more, full of bash code.
https://github.com/nkh/ftl/blob/main/config/ftl/etc/core/ftl

As it is, it's not recognized.

If the extension is changed to .sh, it is recognized as a shell script

if "#!/bin/bash/ is at the beginning of the file, it is recognized as bash code

And now that I runt the test again, I realized that I was wrong.

file ftl -> ftl: Unicode text, UTF-8 text
mimetype ftl -> ftl:   text/plain

but it's right with a shebang:
file  ftl_shebang -> ftl_shebang: Bourne-Again shell script, Unicode text, UTF-8 text executable
mimetype ftl_shebang -> ftl_shebang: application/x-shellscript

I must have mixed files earlier, sorry. 

But let's not lose a good opportunity, I know what those files are, I can cheat and have a list of files and create temporary files, with extension or shebang to give to scc. Or scc could accept a list, and in the best of worlds also generate a list of the files it checked and what types it thought they were.

I'd understand if you feel that the input file with file types (and the list of files/types) is not something you want to implement., I can write a workaround.

boyter:
Ah ok.

So the way `scc` works internally is to check the extension. If its a singular known file type it treats it as that. Where there are multiple it will inspect the first few thousand bytes counting keywords trying to identify the most likely type then count on that.

Where the filename itself matches, such as `makefile` the above applies.

Where nothing matches the file is then checked for the presence of a #! operator.

So what I get from the above is you want to do a remap? This currently exists perhaps.

Have a look at the following options, which might do what you are expecting.

```
--remap-all
--remap-unknown
```

I suspect either of those should work if you are prepared to add a small comment on the top of your files. Although I understand this might not be ideal.

I don't know if any other option is a good idea in this case, at least without it being an opt-in to ensure that performance is not tanked.

nkh:
Thank for pointing at the remapping in this specific case I could add a vim tag to the bash files. I also ran a test with **--remap-all** that worked well (I ned to check the results a bit more).

I have symlinks in the directory structure, that completely broke scc I think as it never finished working.

[...]

PR #694, opened 2026-04-09
Issue #396 Circular Symlinks Bug

RichardSimison:
In issue #396 a user described an issue where scc will hang indefinitely if it is used on a directory containing circular symlink paths. This fixed that by storing scanned directories in a map and skips previously scanned directories if they are encountered again, allowing the program to finish running. This fix also includes tests for broken symlinks that do not point to a real directory, as well as making sure directories are not double counted.


boyter:
Neat. Thanks for this.

PR #695, opened 2026-04-12
fix: prevent apostrophes in docstrings from breaking parsing

### LINKED ISSUE #246: Python docstrings mishandle apostrophes

lawrence3699:
Fixes #246

**Problem:** `docStringState` used `stringTrie.Match()` to detect the end of a docstring. The trie matches *any* string delimiter, so an apostrophe `'` inside a `"""` docstring was incorrectly matched as a string boundary, causing the state machine to exit the docstring prematurely.

**Before:**
```
$ scc test.py   # file contains """ docstring with apostrophe
Python  1  18  1  6  11  0
```
Lines inside the docstring were counted as code.

**After:**
```
$ scc test.py
Python  1  18  2  13  3  1
```
Docstring lines are correctly counted as comments.

**Fix:** Replace `stringTrie.Match()` with `checkForMatchSingle()`, which checks specifically for the `endString` that started the current docstring (e.g. `"""`), rather than matching any string delimiter.

**Validation:** All existing tests pass (`go test ./...`). Updated `TestCountStatsIssue123` expectations (which previously encoded the buggy behavior) and added `TestCountStatsIssue246` regression test.

I explicitly license this contribution under the MIT licence.

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -172,7 +172,7 @@ func stringState(fileJob *FileJob, index int, endPoint int, endString []byte, cu
 
 // This is a special state check pretty much only ever used by Python codebases
 // but potentially it could be expanded to deal with other types
-func docStringState(fileJob *FileJob, index int, endPoint int, stringTrie *Trie, endString []byte, currentState int64) (int, int64) {
+func docStringState(fileJob *FileJob, index int, endPoint int, endString []byte, currentState int64) (int, int64) {
 	// It's not possible to enter this state without checking at least 1 byte so it is safe to check -1 here
 	// without checking if it is out of bounds first
 	for i := index; i < endPoint; i++ {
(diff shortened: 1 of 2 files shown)


boyter:
Neat! Thank you! This is one of those issues I have been meaning to fix for a while.

PR #697, opened 2026-04-13
fix: fix file tests

apocelipes:
#694 causes too many test fail:
image

image

Fix them and do some clean up:

- Clean up `visitedPaths` map when necessary
- Fix tests, especially on macOS, where temporary file directories are symbolic links.
- Move symlink integration tests to go tests


boyter:
Thats on me, I didnt run them. Thanks for fixing.

I was going to do it later tonight.

Issue #412, opened 2023-10-13
Wrong language Complexity/Lines calculation when using -wide


RobertZickler:
**Describe the bug**

When output is generated with `-wide` the `Complexity/Lines` is calculated. This is okay for one ore multiple files. But then the `Complexity/Lines` is accumulated to sum up to the language `Complexity/Lines`. But this is not correct. 

Example:
```
─────────────────────────────────────────────────────────────────────────────────────────────────────────────
Language                              Files     Lines   Blanks  Comments     Code Complexity Complexity/Lines
─────────────────────────────────────────────────────────────────────────────────────────────────────────────
C Header                                  3       526      067        71      338         27            44.70
─────────────────────────────────────────────────────────────────────────────────────────────────────────────
~\the\path\to\project\files\header_file_1.h       397       42        43      312         13             4.17
~\the\path\to\project\files\header_file_2.h       107       21        23       63         11            17.46
~\the\path\to\project\files\header_file_3.h        22        4         5       13          3            23.08
```

$$ 4.17\\,{Complexity \over Lines} + 17.46 \\,{Complexity \over Lines}  + 23.08\\,{Complexity  \over Lines} = 44.70\\,{Complexity  \over Lines} $$

**To Reproduce**

`scc.exe --no-cocomo --no-size --sort complexity --wide header_file_1.h header_file_2.h header_file_3.h`

The error is also produced when  run without `--by-file`

**Expected behavior**

The `Complexity/Lines` should not be added from each file to produce the language-`Complexity/Lines`. In stead it should be calculated on the summed up language Complexity and Lines. In the above example the correct `Complexity/Lines` for C Header would be:

$$ {27 \\,Complexity \over 338 \\,(Code)Lines} = 7.99 \\,{Complexity  \over Lines} $$

_Alternatively: if the described output is by design, the headline `Complexity/Lines` is wrong for the language. But I could not find any description on the reason why the sum of the file `Complexity/Lines` is equivalent to the language `Complexity/Lines`._

One more thing: It would be nice to have the option to sort the files by `Complexity/Lines`.

**Desktop:**
 - OS: Windows
 - Version: scc version 3.1.0

boyter:
https://github.com/boyter/scc/issues/370

boyter:
Thanks for the above... I just fixed it then https://github.com/boyter/scc/pull/719

You nailed the exact issue... a dumb move from me years ago, and since I rarely use wide I never noticed it.

boyter:
[...]

Issue #370, opened 2023-01-25
option -w Complexity/Lines seems odd


astrodidge:
I made some tests with scc (still new to it) and got the following output using the -w option:

Language                              Files     Lines   Blanks  Comments     Code Complexity Complexity/Lines
───────────────────────────────────────────────────────────Go                                       14      3970      595       985     2390        128           126.86
YAML                                      1        16        0         0       16          0             0.00
License                                   1        21        4         0       17          0             0.00
Markdown                                  1       400       82         0      318          0             0.00
─────────────────────────────────────────────────────────────────────────────────────────────────────────────
Total                                    17      4407      681       985     2741        128           126.86

I expected the Complexity/Lines for the Go-Files to be 128 / 2390 * 100 (=5.36) not 126.86

Did I misunderstand the Complexity/Lines figure?

boyter:
You are probably correct. Its one of those options I don't use very often. If you want to investigate it and supply the fix feel free, (you will get the glory of the fix too!) otherwise I will have a look when I get around to it in a week or two.

astrodidge:
thanks for answering my question. Sorry to say that I currently don't have the resources to investigate it. Anyway, a fix is not pressing for me. Feel free to look at the issue any time it suits you well.

boyter:
No problem. I will have a look soon.

boyter:
Should be fixed by #719

PR #700, opened 2026-04-22
Issue #412

vrajpatel-bsu:
Issue: As of this status report scc issue #412 has been fixed. The issue originally was brought up by another user and is listed as issue #370. Both of these issues identify a problem with how Complexity/Lines is calculated when using a parameter called –wide. What should happen when calculating Complexity/Lines is the total complexity and total code lines between all files should be summed in their respective categories and then the Complexity/Lines equation should be used to determine the total complexity of the project. For example Files 1, 2, and 3 have complexity of 10 and code lines of 20. In total the project will have a complexity of 30 and code lines of 60. When the equation is implemented the result should be 50, but it comes out as 150.

Solution: As a solution to this I changed the parameters for Complexity/Lines from summing each of the file’s Complexity/Lines to recalculating the Complexity/Lines from the total complexity and total code lines.  

Test: As a test case for this I created 3 temp files which will have place holder values for each category and make sure the expected total for the Complexity/Lines is returned. Else it will fail and return the entire table to trouble shoot.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/boyter/gocodewalker v1.5.2-0.20260227212453-19676720409f
 	github.com/boyter/simplecache v0.0.0-20250113230110-8a4c9201822a
 	github.com/json-iterator/go v1.1.12
+	github.com/mark3labs/mcp-go v0.45.0
 	github.com/mattn/go-runewidth v0.0.19
 	github.com/rs/zerolog v1.30.0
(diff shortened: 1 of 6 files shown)


apocelipes:
Some commits are duplicated for #694. Only these 3 are new:

image

Maybe the branch should be updated. @vrajpatel-bsu @boyter

boyter:
Going to need the above resolved before I consider merging sorry.

PR #701, opened 2026-04-23
feat(rust): count ? try operator in complexity

daulet:
```
  fn parse<T: ?Sized>(s: &str) -> Result<i32, Error> {
      let n = s.trim().parse()?;
      if n > 0 {
          Ok(n + lookup(n)?)
      } else {
          Err(Error::Negative)
      }
  }

  ┌────────┬─────────────────────────────────────┐
  │        │             Complexity              │
  ├────────┼─────────────────────────────────────┤
  │ before │ 2 — if, else                        │
  ├────────┼─────────────────────────────────────┤
  │ after  │ 4 — if, else, ? (parse), ? (lookup) │
  └────────┴─────────────────────────────────────┘
```

### DIFF
--- a/processor/workers.go
+++ b/processor/workers.go
@@ -110,26 +110,35 @@ func nextNonWhitespaceIndex(content []byte, index int) int {
 	return index
 }
 
-func isRustRelaxedSizedBound(content []byte, index, offsetJump int) bool {
-	if content[index] != '?' {
-		return false
+func hasPostfixExclude(content []byte, index, offsetJump int, excludes [][]byte) bool {
+	for _, exclude := range excludes {
+		if len(exclude) < offsetJump || !bytes.Equal(content[index:index+offsetJump], exclude[:offsetJump]) {
+			continue
+		}
+
+		remaining := exclude[offsetJump:]
+		next := nextNonWhitespaceIndex(content, index+offsetJump)
+		if next+len(remaining) > len(content) || !bytes.Equal(content[next:next+len(remaining)], remaining) {
+			continue
+		}
+
+		afterExclude := next + len(remaining)
+		if len(remaining) != 0 && isIdentifierContinue(remaining[len(remaining)-1]) {
+			return afterExclude == len(content) || !isIdentifierContinue(content[afterExclude])
+		}
+
+		return true
 	}
 
-	next := nextNonWhitespaceIndex(content, index+offsetJump)
-	if next+len("Sized") > len(content) || string(content[next:next+len("Sized")]) != "Sized" {
-		return false
-	}
-
-	afterSized := next + len("Sized")
-	return afterSized == len(content) || !isIdentifierContinue(content[afterSized])
+	return false
 }
 
-func countComplexityPostfix(fileJob *FileJob, index, offsetJump int) {
+func countComplexityPostfix(fileJob *FileJob, index, offsetJump int, langFeatures LanguageFeature) {
 	if !hasNonWhitespaceBefore(fileJob.Content, index) {
 		return
 	}
 
-	if fileJob.Language == "Rust" && isRustRelaxedSizedBound(fileJob.Content, index, offsetJump) {
+	if hasPostfixExclude(fileJob.Content, index, offsetJump, langFeatures.PostfixExcludes) {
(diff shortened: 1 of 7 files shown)


processor/workers.go
```diff
@@ -286,6 +294,15 @@ func codeState(
...
 					fileJob.ComplexityLine[len(fileJob.ComplexityLine)-1] = fileJob.ComplexityLine[len(fileJob.ComplexityLine)-1] + 1
 				}
+
+			case TComplexityPostfix:
+				if index > 0 && !isWhitespace(fileJob.Content[index-1]) {
+					next := index + offsetJump
+					if next >= len(fileJob.Content) || !isIdentifierStart(fileJob.Content[next]) {
+						fileJob.Complexity++
+						fileJob.ComplexityLine[len(fileJob.ComplexityLine)-1] = fileJob.ComplexityLine[len(fileJob.ComplexityLine)-1] + 1
+					}
```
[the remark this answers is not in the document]
daulet: Thanks, agreed. foo() ?; is valid Rust, so the immediate-preceding-byte check does undercount try propagation when whitespace appears before ?.

I updated the matcher to skip backward over whitespace and require a previous non-whitespace byte. I did not rely on a general next-byte identifier check because valid try expressions can be followed by identifiers after whitespace, for example foo()? as u16. Instead, the next-token check is Rust-specific and only suppresses the relaxed ?Sized trait-bound form, including T:?Sized, T: ?Sized, and T:? Sized.

 Added tests for spaced try, ? followed by as, and the ?Sized cases. go test ./... passes.

processor/workers.go
```diff
@@ -88,6 +88,55 @@ func isWhitespace(currentByte byte) bool {
...
+	return index
+}
+
+func isRustRelaxedSizedBound(content []byte, index, offsetJump int) bool {
+	if content[index] != '?' {
+		return false
+	}
+
+	next := nextNonWhitespaceIndex(content, index+offsetJump)
+	if next+len("Sized") > len(content) || string(content[next:next+len("Sized")]) != "Sized" {
```
apocelipes: Maybe you should use `bytes.Compare` here.

processor/workers.go
```diff
@@ -88,6 +88,55 @@ func isWhitespace(currentByte byte) bool {
...
+	afterSized := next + len("Sized")
+	return afterSized == len(content) || !isIdentifierContinue(content[afterSized])
+}
+
+func countComplexityPostfix(fileJob *FileJob, index, offsetJump int) {
+	if !hasNonWhitespaceBefore(fileJob.Content, index) {
+		return
+	}
+
+	if fileJob.Language == "Rust" && isRustRelaxedSizedBound(fileJob.Content, index, offsetJump) {
```
apocelipes: Other languages have same issues, I think hard coded "Rust" is not the solution. For example, we use `name?` to represent an optional property in TypeScript:

```typescript
interface User {
  name: string
  age?: number
}

obj1.attr1?.method1()
```

apocelipes (changes requested):

processor/workers.go
daulet: you want me to update typescript with new field too?

apocelipes: I hope you can implement a general solution. Seems like you have already implemented it.

processor/workers.go
```diff
@@ -88,6 +88,64 @@ func isWhitespace(currentByte byte) bool {
...
+
+	return false
+}
+
+func countComplexityPostfix(fileJob *FileJob, index, offsetJump int, langFeatures LanguageFeature) {
+	if !hasNonWhitespaceBefore(fileJob.Content, index) {
+		return
+	}
+
+	if hasPostfixExclude(fileJob.Content, index, offsetJump, langFeatures.PostfixExcludes) {
```
apocelipes: As a performance hint, it's better to check `len(langFeatures.PostfixExcludes) > 0` before calling the function. Because `hasPostfixExclude` cannot be inlined, calling it on the hot path slows scc down.

boyter:
Thanks for this. This is one of those ones I have wanted for a while but never tackled... probably because I dont do much rust.

PR #703, opened 2026-05-02
Add Move language support (Sui, Aptos, Movement)

poelzi:
## Summary

Adds support for the [Move](https://move-language.github.io/move/) smart-contract language, used by Sui, Aptos, Starcoin, and Movement. `.move` files are currently silently ignored by scc.

A single language definition covers every dialect — Move 2 (Aptos) is a lexical superset of core Move, and Sui only adds the `entry` modifier; no dialect renames or removes a keyword in a way that would confuse a tokenizer.

Notable choices:
- `nestedmultiline: true` — Move's lexer (`move-compiler/src/parser/lexer.rs`) explicitly tracks nested `/* /* */ */` comments via a stack, like Rust.
- Complexity checks mirror the existing Rust entry plus `match` (Move 2) and `abort ` — Move's panic primitive, used on every error path.
- `b"..."` and `x"..."` byte/hex string prefixes work transparently because string parsing begins at the `"`.

## Test plan

- [x] `go generate ./...` regenerates `processor/constants.go` with the new entry
- [x] `go build` compiles cleanly
- [x] `go test ./...` passes (including `TestSpecificLanguages`, which greps CSV output of `examples/language/`)
- [x] `scc examples/language/move.move` reports correct counts: 32 lines, 3 comments (nested block correctly counted as one region), complexity 8

### DIFF
--- a/examples/language/move.move
+++ b/examples/language/move.move
@@ -1,0 +1,32 @@
+module example::counter {
+    use std::signer;
+
+    /// A simple counter resource.
+    struct Counter has key {
+        value: u64,
+    }
+
+    /* outer
+       /* nested */ block */
+    public entry fun increment(account: &signer) acquires Counter {
+        let addr = signer::address_of(account);
+        let counter = borrow_global_mut<Counter>(addr);
+        if (counter.value == 0) {
+            counter.value = 1;
+        } else {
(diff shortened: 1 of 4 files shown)


boyter:
Thanks for this. Will make the next release.

PR #704, opened 2026-05-06
Issue#462

RichardSimison:
I created a –languages-file flag that allows users to use their own custom languages.json file to replace or augment the default json without altering the original code. The implementation involved adding a LanguagesFile global variable in processor.go, registering the flag in main.go, and adding logic at the start of ProcessConstants to read and parse the external JSON file. This merges the custom language definitions into the existing languages map of scc. 

To use the feature, create a JSON file following the same structure as the built-in languages.json and pass it via the flag EX:  scc --languages-file custom.json, where entries with existing language names will override the built-in definition and new names will be added as entirely new languages.

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -3,6 +3,7 @@
 package processor
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -145,6 +146,9 @@ var RemapUnknown = ""
 
 // RemapAll allows remapping of all files with a string to search the content for
 var RemapAll = ""
+
+// LanguagesFile is the path to a custom languages.json file
+var LanguagesFile = ""
 
 type remapRule struct {
 	pattern  []byte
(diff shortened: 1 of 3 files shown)


processor/processor.go
```diff
@@ -324,6 +328,21 @@ func ConfigureLazy(lazy bool) {
...
+			var customLanguages map[string]Language
+			if err := json.Unmarshal(data, &customLanguages); err != nil {
+				printError("failed to parse languages file: " + err.Error())
+			} else {
+				for name, lang := range customLanguages {
+					languageDatabase[name] = lang
+				}
+			}
+		}
+	}
```
boyter, line 345: I think... this needs to come after the load, because we want to allow overriding of existing config. Part of the point being someone may want to count docstrings as code for example.

This also would allow testing differences between configs a little more easily.

boyter (changes requested):
One small fix requested, which I think should be pretty easy.

boyter:
@RichardSimison not sure if you want to fix this... I am ready and willing to merge it in, but I need a rebase and apply on top (should be pretty easy to do). Otherwise ill be reimplementing sometime in the next month.

PR #705, opened 2026-05-08
feat: support recognizing .diff files as Patch

apocelipes:
".diff" and ".patch" files are often used as input for the `patch` command or `git apply`, so they are treated as similar types. Cloc uses a similar classification as well.


boyter:
Chur

PR #711, opened 2026-06-07
feat: little performance tweaks

apocelipes:
Eliminate unnecessary type conversions and use the new, more efficient API.


boyter:
Bound to be a few with all the git stuff I added. I was more focussed on getting it working than performance due to the overhead of it running. Cheers.

PR #712, opened 2026-06-10
refactor: apply go fix suggestions

alexandear:
This PR modernizes the codebase using `go fix -stringsbuilder=false ./...`

See https://go.dev/blog/gofix


boyter:
Thanks for that.

PR #727, opened 2026-06-26
fix: fix tests data race

apocelipes:
Some parallel tests cause data races because they accidentally modify global variables. You can reproduce the issue using the following command:

```console
go test -race -count=50 -run 'TestPreScanConfig|TestRegisterFlagsSliceDefValue|TestRegisterFlagsExhaustive|TestRegressionScalarDefaultsPreserved|TestRegressionShorthandsPreserved' .
```

These tests themselves do not rely on global state, so fix them so that they do not affect global variables.

### DIFF
--- a/SCC-OUTPUT-REPORT.html
+++ b/SCC-OUTPUT-REPORT.html
@@ -18,7 +18,7 @@
 		<th>2131</th>
 		<th>34957</th>
 		<th>4478</th>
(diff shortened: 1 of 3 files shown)


boyter:
Ah thats what it was. I was unable to replicate on my laptop. Cheers.

PR #729, opened 2026-07-04
Cognitive Complexity

boyter:
WIP adding cognative complexity... well an approximation of it.

We want to track cognitive complexity nesting. Cognitive complexity
means we assign higher complexity to nested branch conditions, so

```
if something:
	  if otherthing:
```

would be assigned a higher complexity than

```
if something:
if otherthing:
```

because the nested if requires more mental overhead. To do this we need to track
how nested each condition is when we hit it. We do this by counting the number of
whitespace characters are in front of the condition.
This is an appoximation, true for languages like Python, and probably true for anything
else. However, the benefit of this approach is that it's almost free from a CPU point of view
and the increase in spotting complex code, is genuinely useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ features listed below may be missing from your installation.
 ```text
 $ scc -h
 Sloc, Cloc and Code. Count lines of code in a directory with complexity estimation.
-Version 3.8.0 (beta)
+Version 4.0.0 (beta)
 Ben Boyter <ben@boyter.org> + Contributors
 
 Usage:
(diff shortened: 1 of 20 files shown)


boyter:
```
$ hyperfine 'scc' 'scc --cognitive'
Benchmark 1: scc
  Time (mean ± σ):      4.573 s ±  0.047 s    [User: 11.815 s, System: 6.274 s]
  Range (min … max):    4.520 s …  4.658 s    10 runs

Benchmark 2: scc --cognitive
  Time (mean ± σ):      4.583 s ±  0.086 s    [User: 12.069 s, System: 6.188 s]
  Range (min … max):    4.521 s …  4.816 s    10 runs

Summary
  scc ran
    1.00 ± 0.02 times faster than scc --cognitive
```

No performance regression from this, and the impact of it is actually almost nothing when turned on. I was expecting 5% overhead, but its closer to benchmark noise.

PR #730, opened 2026-07-07
Limit Docker container permissions

ReactorScram:
Use `:ro` on the volume and `--network none` so the container can't write to disk or make any network requests. Users can turn this off if they need something advanced, but it works for me for default invocations.


boyter:
Nice. Approved.

PR #734, opened 2026-07-26
feat: update WGSL and WESL language information

apocelipes:
Also added a WGSL example.


boyter:
TIL they do nested comments.

PR #737, opened 2026-08-08
feat: add F# string literals so (* inside a string is not a comment opener

YuriNachos:
Fixes scc mis-counting F# files where a `(*` inside a string literal opens a phantom multi-line comment.

## Summary
- Add F# string literals to `languages.json` (`quotes` was empty): regular `"…"` (escaped), verbatim `@"…"` (no escape), and triple-quoted `"""…"""` (F# 5+, no escape, may span lines).
- Regenerate `processor/constants.go` and `LANGUAGES.md` via `go generate`.
- Add `TestCountStatsFSharpStringHidesBlockComment` proving a `(*` inside a string no longer eats the following code lines.

## Root Cause
F# declares `(* … *)` as a multi-line comment but had an empty `quotes` array. scc's scanner only ignores comment delimiters while inside a declared string, so a `(*` inside any `"…"` string opened a real block comment that consumed every subsequent line as a comment until EOF or a stray `*)`. A 5-line fixture reported `Code=1, Comment=4` instead of `Code=4, Comment=1`.

F# string syntax per the [F# language reference](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/literals#strings): regular `"…"`, verbatim `@"…"`, and triple-quoted `"""…"""`. Modelled on the existing C# `@"…"` and Boo/Cangjie `"""…"""` entries.

## Validation
- [x] `go generate` (regenerates constants.go + LANGUAGES.md, no diff)
- [x] `go build ./...`
- [x] `go test -race ./...`
- [x] `go test -run TestCountStatsFSharpStringHidesBlockComment ./processor/ -v` (red on master, green after fix)
- [x] 5-line reproducer now reports `Code=4, Comment=1`

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -2381,7 +2381,22 @@
     "extensions": ["fs", "fsi", "fsx", "fsscript"],
     "line_comment": ["//"],
     "multi_line": [["(*", "*)"]],
-    "quotes": []
+    "quotes": [
+      {
+        "end": "\"",
+        "start": "\""
+      },
+      {
+        "end": "\"",
+        "ignoreEscape": true,
+        "start": "@\""
+      },
+      {
+        "end": "\"\"\"",
+        "ignoreEscape": true,
+        "start": "\"\"\""
+      }
+    ]
   },
   "F*": {
     "complexitychecks": [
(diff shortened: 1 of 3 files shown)


boyter:

PR #739, opened 2026-08-08
Add AL (Microsoft Dynamics 365 Business Central)

SShadowS:
Adds AL, the language of Microsoft Dynamics 365 Business Central.

## The language

AL is what Business Central extensions are written in. It is a recognised language in GitHub Linguist (since 2020), and has editor support in VS Code (Microsoft's own extension), Zed, Sublime Text and Helix.

## The definition

```json
"AL": {
  "complexitychecks": ["if ", "if(", "IF ", "IF(", "case ", "CASE ", "while ", "WHILE ", "for ", "FOR ", "foreach ", "FOREACH ", "until ", "UNTIL ", "else ", "ELSE ", "and ", "AND ", "or ", "OR ", "xor ", "XOR ", "not ", "NOT "],
  "extensions": ["al"],
  "line_comment": ["//"],
  "multi_line": [["/*", "*/"]],
  "quotes": [{"end": "'", "start": "'"}]
},
```

AL uses `//` line comments, `/* */` block comments, and single quoted string literals with an embedded quote written `''`. The complexity keywords are AL's branching and looping constructs, in both cases. AL is case insensitive, and code converted from the older C/AL is written in upper case, so both forms are carried the way Pascal does it.

`processor/constants.go` and `LANGUAGES.md` are regenerated with `go generate ./...`.

## On the extension

`.al` is shared with Perl's AutoLoader, which is worth flagging because it does cause misclassification in other tools. It is not an issue here: nothing in `languages.json` currently claims `al`, so this is a clean addition rather than a reassignment, and no existing count changes.

## Example file and tests

`examples/language/al.al` covers a standalone line comment, a trailing line comment, a multi-line block comment, two branches so the complexity checks are exercised, and three strings that would break a naive counter:

```al
Note := 'It''s at https://example.com';
Note := 'not /* a comment */ either';
Note := '// not a comment either';
```

scc counts it as 35 lines, 4 blank, 4 comment, 27 code, complexity 2, which matches a hand count. The two complexity hits are a lower case `while` and an upper case `IF`, so both keyword cases are exercised. All three strings are counted as code, so neither comment syntax leaks out of a quoted string.

`"AL"` is registered in `main_test.go`, and `TestSpecificLanguages` passes.

The suite reports 9 failures on this machine, all of them present on a clean checkout of `master` as well (they are Windows path and line ending issues in the golden report and coupling tests, unrelated to this change).

`SCC-OUTPUT-REPORT.html` is not regenerated, since `go generate` does not produce it; happy to refresh it if you would like.

## AI disclosure

This change was prepared with AI assistance (Claude). The counts above were verified by hand and by running scc locally.


boyter:

PR #740, opened 2026-08-08
fix: report non-zero megabyte size for xkcd-kb unit in non-leap years

YuriNachos:
The `xkcd-kb` size case assigned `size` only inside the leap-year branch with no `else`, so in a non-leap year `size` stayed `0.0` and `scc` printed `0.000 megabytes (XKCD-KB)` for every file — in 3 of every 4 years. This contradicts the unit's own help text "1000 bytes during leap years, 1024 otherwise" (which says non-leap must be 1024-based = 1,048,576). Reproduced on master: 3,000,001-byte file, year 2026 → "0.000 megabytes". Every other `--size-unit` emits a real figure.

## Fix
Divide by a new pure helper `xkcdKbDivisor(year)` returning `1_000_000` (1000-based) in a leap year and `1_048_576` (1024-based) otherwise — mirroring the existing pure `isLeapYear(year)`. No `languages.json` change, so no `go generate` drift.

## Test plan
`go test -race ./processor/...` — green, including a new `TestCalculateSizeXkcdKb` that asserts the output is non-zero in a non-leap year and matches `10_000_000/(1024*1024) = 9.537 megabytes`, plus year-independent assertions `xkcdKbDivisor(2026) == 1_048_576` and `xkcdKbDivisor(2024) == 1_000_000`.

### DIFF
--- a/processor/formatters_cost.go
+++ b/processor/formatters_cost.go
@@ -144,9 +144,7 @@ func calculateSize(sumBytes int64, str *strings.Builder) {
 		size = float64(sumBytes) / 1_024_000
 	case "xkcd-kb":
 		str.WriteString("1000 bytes during leap years, 1024 otherwise\n")
-		if isLeapYear(time.Now().Year()) {
-			size = float64(sumBytes) / 1_000_000
-		}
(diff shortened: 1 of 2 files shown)


boyter:

PR #742, opened 2026-08-09
Fix --cocomo-project-type panic for case-variant built-in types

YuriNachos:
ProcessConstants lower-cased the project-type for its membership check but the COCOMO indexers used the raw value, so any case-variant built-in (e.g. `Organic`) indexed a nil slice and panicked. Route every projectType lookup through a cocomoCoefficients guard that normalizes to canonical lowercase and falls back to organic.


boyter:
@copilot resolve the merge conflicts in this pull request

boyter:
Cheers

PR #744, opened 2026-08-10
fix(terraform): count string literals so /* inside a glob is not a comment

YuriNachos:
## Problem
`Terraform` (an HCL dialect) declares an empty `quotes` list, unlike the `HCL` language entry which declares the `"..."` string literal. With no string rule, a value such as `"logs/*"` is parsed as code: the embedded `/*` opens a block comment that runs to the next `*/` or to EOF, swallowing every following line as a comment.

```hcl
variable "glob" {
  default = "logs/*"
}
output "result" {
  value = "still real code"
}
```

On master this reports **6 lines / 2 code / 4 comments** — the `}` and the whole `output` block are miscounted as comments. Globs like `"logs/*"` are common in Terraform (`ignore_changes`, `for_each` filters, path variables).

## Fix
Add the single `"..."` quote rule that `HCL` already carries, then regenerate `processor/constants.go`. No other language is touched.

## Test plan
`go test ./...` — green, including a new `TestCountStatsTerraformGlobStringNotComment` that feeds the glob above and asserts `6 lines / 6 code / 0 comments / 0 blanks`. On master the same input returns `4 comments` (the test fails); with the fix it returns `0 comments`.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -7868,7 +7868,12 @@
     "extensions": ["tf", "tfvars", "tf.json"],
     "line_comment": ["#", "//"],
     "multi_line": [["/*", "*/"]],
-    "quotes": []
+    "quotes": [
+      {
+        "end": "\"",
+        "start": "\""
+      }
+    ]
   },
   "Textile": {
     "complexitychecks": [],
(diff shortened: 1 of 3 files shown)


boyter:
Cheers

PR #745, opened 2026-08-10
fix(protobuf): count string literals so /* inside an option is not a comment

YuriNachos:
## Problem
`Protocol Buffers` declares an empty `quotes` list, despite using `"..."` string literals throughout the language — `syntax = "proto3";`, `option java_package = "com.example";`, string field defaults, and validation regex options. With no string rule, an option value such as `"see /* legacy block"` is parsed as code: the embedded `/*` opens a block comment that runs to the next `*/` or to EOF, swallowing every following line as a comment.

```proto
syntax = "proto3";
package demo;

option (desc) = "see /* legacy block";

message Foo {
  string name = 1;
}
```

On master this reports **8 lines / 3 code / 4 comments** — the blank line, `message Foo {`, `string name = 1;` and `}` are all miscounted as comments.

## Fix
Add the single `"..."` quote rule (Protobuf string literals are double-quoted) and regenerate `processor/constants.go`. No other language is touched.

## Test plan
`go test ./...` — green, including a new `TestCountStatsProtobufStringNotComment` that feeds the option-string above and asserts `8 lines / 6 code / 0 comments / 2 blanks`. On master the same input returns `4 comments` (the test fails); with the fix it returns `0 comments`.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -5993,7 +5993,12 @@
     "extensions": ["proto"],
     "line_comment": ["//"],
     "multi_line": [["/*", "*/"]],
-    "quotes": []
+    "quotes": [
+      {
+        "end": "\"",
+        "start": "\""
+      }
+    ]
   },
   "Puppet": {
     "complexitychecks": [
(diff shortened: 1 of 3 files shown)


boyter:
Cheers

PR #746, opened 2026-08-10
fix: guard zero category totals in --percent tabular output

YuriNachos:
## Problem
The wide (`fileSummarizeLong`) and short (`fileSummarizeShort`) `--percent` rows divide each category total with **no zero guard**. A project whose files have no blank/comment/complexity lines therefore prints `NaN%` for those columns:

```
$ scc --percent a.json
JSON               1        1         0         0          1          0
Percentage       100.0%    100.0%      NaN%       NaN%     100.0%       NaN%
```

The sibling JSON path (`addLanguagePercentages`) and the `Complexity/Lines` total both already guard `total != 0` — only the tabular percentage rows were inconsistent.

## Fix
Factor the division into a `pct(value, total)` helper that returns `0` when `total == 0`, matching the JSON behaviour; each empty category now reports `0.0%`.

## Test plan
`go test ./...` — green, including a new `TestPercentNoNaNOnZeroCategory` that feeds a single file with `Blank=Comment=Complexity=0` and asserts the short and wide output contain `0.0%` and no `NaN`. On master the same input emits `NaN%` (the test fails).

### DIFF
--- a/processor/formatters_percent_test.go
+++ b/processor/formatters_percent_test.go
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+
+package processor
+
+import (
+	"strings"
+	"testing"
+)
+
+// A project whose files carry zero blank/comment/complexity lines used to make
+// the tabular --percent output print NaN% for those columns, because the
+// percentage divisions had no zero-total guard (unlike addLanguagePercentages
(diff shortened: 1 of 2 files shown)


boyter:
Odd. I thought I had caught all of those. Thanks.

Issue #510, opened 2024-09-24
when using scc with github actions json2 does not produce json output


usmaniqbal998:
**Describe the bug**
when i use  `scc -f json2 -o scc_results.json` in my github actions i do not get json results instead in a raw text format 

**To Reproduce**

   - name: Run scc
      shell: bash
      id: scc
      working-directory: ${{ github.workspace }}
      run: |
        scc -f json2 -o scc_results.json
        
 This is how i am collectiong scc results . my  am using bash and i have `go-version: "1.20"` 
 
 the output i get is plain text as following 

`───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
TypeScript                 654     42267     4805       723    36739       2702
Sass                        97      4724      681        14     4029          0
JSON                        57     44214        1         0    44213          0
JavaScript                  30      4107      339      1121     2647        140
YAML                        18       725       79        14      632          0
Markdown                     9      1649      645         0     1004          0
SVG                          8        25        0         0       25          0
Shell                        3        18        5         6        7          0
Plain Text                   2        43        5         0       38          0
TypeScript Typings           2        97       17         7       73          3
CSV                          1       428        0         0      428          0
Docker ignore                1        36        2         4       30          0
Dockerfile                   1         8        2         0        6          0
HTML                         1        44        4        13       27          0
License                      1         0        0         0        0          0
gitignore                    1        35        7         5       23          0
───────────────────────────────────────────────────────────────────────────────
Total                      886     98420     6592      1907    89921       2845
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop $3,041,932
Estimated Schedule Effort 23.414646 months
Estimated People Required 15.389219
───────────────────────────────────────────────────────────────────────────────
`

If i switch to `json` instead of `json2` output then i dont get COCOMO estimations

following is full code for my github action 

`name: "SCC Analysis"
description: "Run scc, display results"
inputs:
  fail-threshold:
    description: "Threshold for failing the build based on complexity"
    required: false
    default: 100
outputs:
  complexity:
    description: "The total complexity of the codebase"
runs:
  using: "composite"
  steps:
    - name: Check out repository
      uses: actions/checkout@v3

    - name: Set up Node.js
      uses: actions/setup-node@v3
      with:
        node-version: "20.x"

    - name: Set up Go
      uses: actions/setup-go@v4
      with:
[...]

PR #748, opened 2026-08-15
fix: report a failed -o write instead of claiming success

VXNCXNX:

## What's broken

`-o` reports success after a write that failed, and exits 0.

```
$ scc --format json -o /nonexistentdir/z.json .
results written to /nonexistentdir/z.json
$ echo $?
0
$ ls /nonexistentdir/z.json
ls: cannot access '/nonexistentdir/z.json': No such file or directory
```

`--format-multi` has the same hole, exiting 0 with the message going to stdout:

```
$ scc --format-multi 'json:/nonexistentdir/m.json' . ; echo $?
0
```

The same mistakes on the sibling paths are already handled, which is what makes this look like an oversight:

```
$ scc --hotspots --depth 20 -o /nonexistentdir/x.json . ; echo $?
open /nonexistentdir/x.json: no such file or directory
1
$ scc --report=/nonexistentdir/r.html . ; echo $?
create report file: open /nonexistentdir/r.html: no such file or directory
1
```

Read-only directories and paths that are actually directories fail the same silent way. In a CI job that is a green build with no artifact.

## The fix

`processor.go` had `_ = os.WriteFile(...)` followed unconditionally by the success line. It now checks the error, prints it to stderr and exits 1, following `runReport` a few hundred lines above, which does exactly this. `formatters.go` already built a message for the multi case but sent it to stdout and carried on, so it goes to stderr and exits too.

After:

```
$ scc --format json -o /nonexistentdir/z.json . ; echo $?
open /nonexistentdir/z.json: no such file or directory
1
$ scc --format-multi 'json:/nonexistentdir/m.json' . ; echo $?
/nonexistentdir/m.json unable to be written to for format json: open /nonexistentdir/m.json: no such file or directory
1
```

A successful write is unchanged, still `results written to ...` and exit 0.

`os.Exit(1)` rather than returning an error, because `Process()` is called from cobra's `Run:`, which has no error return, and the report path at `processor.go:932` already resolves that the same way. Happy to plumb an error through instead if you would rather change the signature.

**This changes an exit code**, from 0 to 1, on a case that previously failed silently. That is the point of the change, but it is a visible difference for anyone whose script currently ignores it.

## Verification

`TestRegressionOutputWriteFailureExits` in `regression_test.go`, one subtest per path. It observes the real exit code: `runSCCDir` runs the test binary as a subprocess, so `os.Exit(1)` is a genuine child-process exit. It asserts non-zero exit, absence of the success message, presence of the reason, and that no file was created. With both source changes reverted and the test kept, both subtests fail.

`LANG=C go test ./...` passes.

A note in case you see it too: on a machine with a comma-decimal `LANG`, four unrelated tests fail on master, because the formatters build their printer from `os.Getenv("LANG")` while the fixtures spell numbers with a dot. `LANG=C` is green before and after.

## On issue #510

I looked at whether this explains #510 and I do not think it does. That reporter got the tabular summary written instead of JSON, which is the unknown-format fallback from a `json2`-less binary, matching your reply that the action needed updating. Their target directory was writable, so no write failed and nothing was swallowed. Same family of symptom, different cause.

### DIFF
--- a/processor/formatters.go
+++ b/processor/formatters.go
@@ -196,7 +196,10 @@ func fileSummarizeMulti(input chan *FileJob) string {
 			} else {
 				err := os.WriteFile(t[1], []byte(val), 0600)
 				if err != nil {
-					fmt.Printf("%s unable to be written to for format %s: %s", t[1], t[0], err)
+					// A failed write must not report success: match the -o
+					// path and exit non-zero with the reason on stderr.
+					fmt.Fprintf(os.Stderr, "%s unable to be written to for format %s: %s\n", t[1], t[0], err)
+					os.Exit(1)
 				}
 			}
 		}
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1113,7 +1113,12 @@ func Process() {
 	if FileOutput == "" {
 		fmt.Print(result)
 	} else {
-		_ = os.WriteFile(FileOutput, []byte(result), 0644)
+		// A failed write must not report success: follow the --report path
+		// above and exit non-zero with the reason on stderr.
+		if err := os.WriteFile(FileOutput, []byte(result), 0644); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
 		fmt.Println("results written to " + FileOutput)
 	}
 }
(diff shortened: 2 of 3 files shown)


boyter:
Cheers.

PR #749, opened 2026-08-15
fix(openmetrics): emit parseable OpenMetrics

VXNCXNX:

## What's broken

`--format openmetrics` does not produce parseable OpenMetrics, for three independent reasons.

**1. Metric families are interleaved.** All seven `# TYPE`/`# HELP` blocks are written up front, then each language emits all seven metrics, so the samples of one family are scattered through the output. The spec requires them contiguous.

```
$ scc --format openmetrics . > s.txt
$ python3 -c "from prometheus_client.openmetrics.parser import text_string_to_metric_families as om; list(om(open('s.txt').read()))"
ValueError: Clashing name: scc_files
```

**2. The summary path never writes `# EOF`.** The per-file path does. Without it the parser reports `Missing # EOF at end`.

**3. Label values are not escaped.** A quote in a filename closes the label early:

```
$ cp a.py 'we"ird.py' && scc --format openmetrics --by-file .
scc_lines{language="Python",file="/tmp/.../we"ird.py"} 1
```

The file label escapes backslashes but never quotes, and the language label is not escaped at all. The sibling formatters already handle this: CSV writes `"we""ird.py"`, SQL writes `'it''s.py'`.

## The fix

Both loops now iterate metric family first, then language or file, so each family's metadata is immediately followed by its own samples. The metadata blob became a table pairing each metric with its accessors, which is what makes that ordering possible. The summary path gained the missing `# EOF`, and a small escaper covering `\`, `"` and newline is applied to both label sites.

`toOpenMetricsFiles` now drains the channel into a slice first, since samples of one family can no longer be emitted as each file streams past.

After, on a directory containing `we"ird.py`:

```
--format openmetrics            OK  7 families, 14 samples
--format openmetrics --by-file  OK  7 families, 18 samples
scc_lines{language="Python",file="/tmp/.../we\"ird.py"} 1
```

**The sample values are unchanged.** Comparing sorted non-comment lines before and after: the summary output is byte-identical, and the by-file output differs only in the escaped quotes. Only grouping, the terminator, and escaping changed.

One thing to flag: `scc_files` has no per-file sample, so in `--by-file` its metadata is emitted with zero samples. The parser accepts that, and it matches the previous behaviour where it also had no per-file samples, but it is now visible as a lone `# TYPE`/`# HELP` pair.

## Verification

The three existing tests are exact-equality assertions, so their expected literals are rewritten to the new grouping rather than weakened. `TestFileSummarizeOpenMetricsPerFile` keeps its `C:\\bbbb.go` backslash expectation, which still passes.

Three new tests: two asserting family contiguity and the `# EOF` terminator, one asserting label escaping. The contiguity helper was mutation-checked by temporarily restoring the language-first loop, which fails it with `Samples of family are not contiguous: scc_files`, so it is not vacuous.

`LANG=C go test ./...` passes, and the `prometheus_client` OpenMetrics parser accepts both output modes.

Same locale note as my other PRs: on a comma-decimal `LANG` four unrelated tests fail on master, because the formatters build their printer from `os.Getenv("LANG")`. `LANG=C` is green before and after.

### DIFF
--- a/processor/formatters_misc.go
+++ b/processor/formatters_misc.go
@@ -9,24 +9,91 @@ import (
 	"go.yaml.in/yaml/v2"
 )
 
-var openMetricsMetadata = `# TYPE scc_files gauge
+// openMetricsMetric describes a single OpenMetrics metric family. OpenMetrics requires all
+// samples of a family to be contiguous, so output is generated metric family by metric family
+// rather than language by language.
+type openMetricsMetric struct {
+	name     string
+	metadata string
+	// summary extracts the sample value from a language summary. A nil value means the metric
+	// has no samples in the per-file output.
+	summary func(LanguageSummary) int64
+	file    func(*FileJob) int64
+}
+
+var openMetricsMetrics = []openMetricsMetric{
+	{
+		name: "files",
+		metadata: `# TYPE scc_files gauge
 # HELP scc_files Number of sourcecode files.
-# TYPE scc_lines gauge
+`,
+		summary: func(l LanguageSummary) int64 { return l.Count },
+	},
+	{
+		name: "lines",
+		metadata: `# TYPE scc_lines gauge
 # HELP scc_lines Number of lines.
-# TYPE scc_code gauge
+`,
+		summary: func(l LanguageSummary) int64 { return l.Lines },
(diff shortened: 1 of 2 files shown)


boyter:
Cheers for this.

PR #751, opened 2026-08-17
fix(lua): long string start and end brackets are swapped

VXNCXNX:
## What's broken

Lua, Luau and XMake declare their long-bracket string with `start` and `end` swapped in `languages.json`:

```json
{ "end": "[[", "start": "]]", "ignoreEscape": true }
```

So `[[` never opens a string. Two consequences, in opposite directions:

- a `--` inside a long string is counted as a comment
- a bare `]]` in ordinary code, like the nested index `data[keys[1]]`, opens a bogus string that swallows every following line until the next `[[` or EOF

`Factor` in the same file has the correct orientation, which is what makes the mistake visible.

## Repro

A file with a long string containing `--`, then `data[keys[1]]`, then one real comment:

```
                Lines  Blanks  Comments  Code
before   Lua        9       0         0     9
after    Lua        9       0         1     8
```

The real comment was invisible because the earlier `]]` had opened a string that never closed.

## The fix

Swap the two fields on those three languages, then `go generate ./...` to regenerate `processor/constants.go`.

## Verification

Three tests. Two assert the fixed behaviour across all three languages, and the third asserts that `--[[ ... ]]` block comments are still counted as comments, which is the guard against over-correcting. Swapping the brackets back in the generated constants fails the first two, and deliberately not the third.

`go test ./...` shows only the 5 failures already present on a clean `master` (`TestIssue457`, both cocomo tests, `TestPercentNoNaNOnZeroCategory`, `TestFileSummarizeLongWeightedComplexity`). I re-ran `go generate` myself and confirmed the committed constants match byte for byte, so the generate check stays green.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -4360,8 +4360,8 @@
         "start": "'"
       },
       {
-        "end": "[[",
-        "start": "]]",
+        "end": "]]",
+        "start": "[[",
         "ignoreEscape": true
       }
     ],
@@ -4410,8 +4410,8 @@
         "start": "`"
       },
       {
-        "end": "[[",
-        "start": "]]",
+        "end": "]]",
+        "start": "[[",
         "ignoreEscape": true
       }
     ],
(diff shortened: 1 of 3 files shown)


boyter:
No idea how I missed that one. Thanks.

PR #752, opened 2026-08-17
fix: correct the block comment delimiters for MATLAB and Racket

NotAFlightRisk:
A `%{ ... %}` block in a MATLAB file swallows everything after it. The closer is listed as `}%`, so the opener never matches and the scanner runs the comment to EOF - a nine-line function counts 1 code / 7 comments / 1 blank, where it should be 3 / 4 / 2.

Racket is the same thing from the other side, its pair stored backwards as `|#` then `#|`. Lisp and Scheme sit in the same file with it the right way round, which is what convinced me it wasnt deliberate.

Nothing moves in `examples/language` either way - `racket.rkt` has no block comments and theres no MATLAB file in there to check against.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -4495,7 +4495,7 @@
       "fprintf("
     ],
     "line_comment": ["%"],
-    "multi_line": [["%{", "}%"]],
+    "multi_line": [["%{", "%}"]],
     "quotes": [
       {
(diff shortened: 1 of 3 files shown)


boyter:
TIL

Issue #464, opened 2024-05-20
Feature: Add option to count files not supported by scc


boyter:
An inverse check which counts them as text files. This would help with identifying odd files that are not binary. Either to identify future languages to support or just to assist with metrics. 

Would put them under a "Unknown" category.

PR #754, opened 2026-08-18
Add --count-unsupported to count unrecognised files as Unknown

ChrisJr404:
This adds a `--count-unsupported` flag that counts files whose language scc doesn't recognise under an "Unknown" category as plain text, rather than skipping them silently, as requested in #464.

By default nothing changes: unrecognised files are still skipped with the usual warning. When `--count-unsupported` is set, a file that detection can't place is labelled `Unknown` and counted like plain text. The `Unknown` category deliberately carries no language features, so no comments or complexity are attributed to it, and the counter's null-byte check still runs, so anything that turns out to be binary is dropped exactly as before. The net effect is that only genuine non-binary text files surface, which makes it easy to spot odd files that might warrant a new language definition or that were quietly missing from the metrics.

The change slots into `newFileJob`: the "unknown extension" path now either skips (default) or relabels to `Unknown` and falls through the existing allow/exclude filters, so the flag composes with `--include-ext`, `--exclude-ext`, and friends. `CountStats` already substitutes empty tries when a language has no registered features, so `Unknown` needs no special-casing downstream.

Testing: added `TestNewFileJobUnsupportedSkippedByDefault` and `TestNewFileJobUnsupportedCounted` (the latter also runs the file through `CountStats` to confirm it counts as plain text with zero comments/complexity). Full `go test ./...` passes, plus a manual run over a mixed directory confirming a text file with an unknown extension lands under `Unknown` while a binary file with an unknown extension is still skipped.

I license this contribution under the MIT licence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,7 @@ Flags:
       --count-as string                     count extension as language [e.g. jsp:htm,chead:"C Header" maps extension jsp to html and chead to C Header]
       --count-as-pattern stringArray        count files matching a path pattern as a new named category backed by a base language [repeatable; pattern is glob by default, prefix with re: for regex; e.g. *_spec.rb:"Ruby Spec":Ruby or re:\.test\.js$:"JavaScript Tests":JavaScript]
       --count-ignore                        set to allow .gitignore and .ignore files to be counted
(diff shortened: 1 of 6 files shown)


boyter:
Neat. Thanks for that. Just in time for the next release too.

PR #755, opened 2026-08-18
fix(ts): correctly detect QT translation source files

serkonda7:
> This PR was assisted by LLM.

QT Translation source files also use the `.ts` extension but are actually xml-based. [^1] [^2]
These are currently misclassified as TypeScript

This PR adds `Qt Translation Source` as language with proper heuristics.

### Testing
Test included.
I also manually verified detection on https://github.com/keepassxreboot/keepassxc and a repo using TypeScript.

[^1]: https://support.phrase.com/hc/en-us/articles/6111346553628--TS-Qt-Translation-Source-Strings
[^2]: https://wiki.qt.io/Qt_Localization


boyter:
Neat. Ill try some tests to ensure not a regression to performance that could be solved with keywords and possibly modify this, but taking it now.

serkonda7:
Thanks! Checking for performance regressions is a good idea

boyter:
Seems like less than 1% so nothing to worry about. Thanks for the submission.

PR #756, opened 2026-08-18
fix: escape language names and file paths in html output

VXNCXNX:
`--format html` and `--format html-table` interpolate file paths and language names into table cells with no escaping, so anything containing `<`, `>`, `&` or `"` breaks out of the cell and produces invalid HTML.

With files named `a<b>.go` and `b".py`:

```
$ scc --by-file --format html-table
before:  <td>/tmp/scctest/a<b>.go</td>
         <td>/tmp/scctest/b".py</td>
after:   <td>/tmp/scctest/a&lt;b&gt;.go</td>
         <td>/tmp/scctest/b&#34;.py</td>
```

`--count-as-pattern` category names reach the header cells the same way:

```
$ scc --count-as-pattern '*.go:<script>alert(1)</script>:Go' --format html-table
before:  <th><script>alert(1)</script></th>
after:   <th>&lt;script&gt;alert(1)&lt;/script&gt;</th>
```

That matters because the html output is a file people open in a browser, and on CI the paths and category names can come from the repository being scanned.

Two `html.EscapeString` calls in `toHtmlTable`. The numeric columns are `%d` so they need nothing. `toHtml` wraps the same function and is covered by the change.

AI-assisted: written with Claude Code, reviewed and tested by me before sending.

### DIFF
--- a/processor/formatters_misc.go
+++ b/processor/formatters_misc.go
@@ -4,6 +4,7 @@ package processor
 
 import (
 	"fmt"
+	"html"
 	"strings"
 
 	"go.yaml.in/yaml/v2"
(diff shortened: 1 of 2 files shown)


boyter:
Good ol edge cases.

PR #758, opened 2026-08-21
feat: improve qt translation source detection

apocelipes:


boyter:
Dammit.. just finished profiling the other one :) Ill get this in and test again. Last one was only < 1% impact, ill see what this does.

Cheers for this.

PR #759, opened 2026-08-24
fix(php): correct the double quoted string delimiter

NotAFlightRisk:
A 13 line PHP file with `glob("$root/modules/*/config.php")` in it comes back from scc 4.0.0 as 8 comments and 4 code. Theres no comment in it anywhere. The same file with single quotes round the glob gives the right 0 and 10.

`languages.json` has PHP's double quote as `\"` rather than `"`, so the scanner never enters a string and the `/*` in the glob opens a block comment that runs to the end of teh file. [The manual](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.double) has the delimiter as a plain `"`, the backslash form is the escape you write inside one.

Left the other 38 entries with an escaped quote alone, they cant all just be unescaped. Vim Script starts a line comment with `"`, so making that a string opener loses both comment lines in `examples/language/vim.vim`.

A heredoc body with an odd number of `"` in it now keeps the string open past the terminator and swallows the comment after it. `'` already behaves that way on master - a heredoc containing `don't` counts the next comment line as code today.


boyter:
I guess a minor release is incoming.

PR #761, opened 2026-08-26
Issue #760 - write response body before returning from the handler

### LINKED ISSUE #760: `cmd/badges/main.go` `filterBad(loc)` logs error and returns without setting a status code or res body
**Describe the bug**
In the badges server (`cmd/badges/main.go`), when an incoming request matches the bad-pattern criteria in `filterBad(loc)`, the HTTP handler logs the error and returns immediately without setting an HTTP status code or writing a response body. This causes the server to return an empty default `200 OK` response with 0 bytes instead of an explicit error status code like `400 Bad Request`.

**To Reproduce**
1. Run the badges server: `go run cmd/badges/main.go`
2. Send a request: `curl -i http://localhost:8080/github/wp-content.com/wp-admin.com`
```
HTTP/1.1 200 OK
Date: Wed, 26 Aug 2026 06:14:09 GMT
Content-Length: 0
```

**Expected behavior**
The handler should set an explicit HTTP error status code (such as `400 Bad Request`) and return an error message in the response body before exiting.

**Proposed Solution**
Write an HTTP 400 Bad Request header and response body before returning from the handler.

**Desktop:**
 - OS: Windows
 - Version: Windows 11

HuanUnited:
fix: #760

### DIFF
--- a/cmd/badges/main.go
+++ b/cmd/badges/main.go
@@ -63,6 +63,8 @@ func main() {
 
 		if filterBad(loc) {
 			log.Error().Str(uniqueCode, "bfee4bd8").Str("loc", loc.String()).Msg("filter bad")
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = w.Write([]byte("invalid request"))
 			return
 		}
 


boyter:
Cheers.

PR #764, opened 2026-08-27
fix: count Rust byte character literals

daulet:
Note: found, fixed and published by 5.6-sol, I mainly reviewed changes for sanity.

Model b'...' as a Rust quote form and advance ordinary multi-byte string prefixes without disabling escape handling. This prevents embedded double quotes from swallowing later comments and complexity tokens.

## Downstream impact

This did not affect Loopy at runtime; it corrupted the source metrics used during review. On Loopy's 7,107-line `src/state.rs`:

| Metric | Before | After |
| --- | ---: | ---: |
| Code | 6,932 | 6,795 |
| Blank | 173 | 310 |
| Complexity | 588 | 1,188 |

The total line count remained unchanged, confirming a parser-state classification error rather than a source-size change.


boyter:
Nice. Thanks for that.

PR #772, opened 2026-09-06
fix(asp.net): correct the server side comment terminator

MaxFreedomPollard:
## What's broken

The ASP.NET server side comment is `<%-- ... --%>`, but `languages.json` line 138 pairs the `<%--` opener with `-->`:

```json
"multi_line": [["<!--", "-->"], ["<%--", "-->"]],
```

so the opener never matches its own closer. `CountStats` enters the block comment state at `<%--`, finds no `-->` after it, and keeps that state to the end of the file, counting every following line as a comment.

`HEEx` in the same file has `["<%!--", "--%>"]`, the same Elixir style tag pair with the right terminator, which is what makes the ASP.NET one look like a slip rather than a decision.

## Repro

`Default.aspx`, nine lines, one single line server side comment on line 2 and no HTML comment anywhere:

```aspx
<%@ Page Language="C#" AutoEventWireup="true" CodeBehind="Default.aspx.cs" Inherits="Demo.Default" %>
<%-- server side comment, stripped before the page is sent --%>
<html>
<body>
    <form id="form1" runat="server">
        <asp:Label ID="Label1" runat="server" Text="Hello" />
    </form>
</body>
</html>
```

```
                     Files  Lines  Blanks  Comments  Code
before   ASP.NET         1      9       0         8     1
after    ASP.NET         1      9       0         1     8
```

Everything from `<html>` down was counted as comment because the comment opened on line 2 and never closed.

## The fix

Change the closer to `--%>` in `languages.json`, then `go generate` to regenerate `processor/constants.go`. The `["<!--", "-->"]` pair in the same entry is untouched, so plain HTML comments in `.aspx` markup count exactly as they did before. `LANGUAGES.md` is regenerated from the same pass and does not change, since it lists names and extensions rather than delimiters.

## Verification

Three tests in `processor/workers_aspnet_comment_test.go`. Two assert the fixed behaviour, a one line `<%-- --%>` comment and a multi line one, and the third is the guard against over-correcting: a `.aspx` file whose only comment is a two line `<!-- -->` still reports 2 comments and 5 code.

Reverting only `languages.json` and `processor/constants.go` to master while keeping the test file, `go test ./processor/ -run TestCountStatsAspNet -v` fails the two behaviour tests with `Expected 8 code got 1` / `Expected 1 comment got 8` and `Expected 2 code got 1` / `Expected 4 comments got 6`, and passes the HTML guard. With the fix applied all three pass.

`go test ./... -p 4` is green on Go 1.26.4 darwin/arm64, and it was green on unmodified `master` before the change too, so there is no pre-existing failure to read around. `gofmt -l .` outside `vendor/` prints nothing and `go vet ./...` is clean. I re-ran `go generate` after committing and `processor/constants.go` came back byte for byte identical, so the generated file matches the JSON.


boyter:
Much appreciated. Will make the next release... which will be sooner than you expect.

MaxFreedomPollard:
Sounds good man

PR #773, opened 2026-09-10
fix(dart): count single quoted string literals

MaxFreedomPollard:
A six line Dart file whose first line is `const iconGlob = 'assets/*/icon.png';` comes back from scc as 1 code, 5 comments, 0 blanks. There is no comment in the file and there is a blank line in it. The `/*` inside the single quoted string opens a block comment that runs to the end of the file, so the blank line and every line after it count as comment.

It bites the other way round too. `const doubleQuote = '"';` followed by a blank line and a real `//` comment gives 6 code and 0 comments, because the `"` inside the single quoted string opens a string that never closes and eats the comment. Same reason, complexity keywords sitting inside a single quoted string get counted, so `const label = 'if you switch it while it is on';` scores 3.

`languages.json` gives Dart one quote form, `"` to `"`. Dart strings are delimited by either quote character, and Dart has no character literal at all, so `'` can only ever be a string delimiter there. Dart's Built-in types page in the language documentation puts it as "You can use either single or double quotes to create a string", and the digit separator on the same page is `_`, so the `'` ambiguity that keeps C and C++ out of this in `processor/workers_charliteral_test.go` does not apply. Single quotes are also the form most real Dart source uses, the linter ships a `prefer_single_quotes` rule for it.

The fix is a second quote entry, `'` to `'`, in the Dart block of `languages.json`, then `go generate` to regenerate `processor/constants.go`. `LANGUAGES.md` comes back unchanged since it lists names and extensions rather than delimiters.

Four tests in `processor/workers_dart_string_test.go`. Three assert the fixed behaviour for the cases above, and the fourth is the guard against over-correcting: a file whose only comments are a real `/* */` block and a real `//` line still reports 3 comments and 3 code, and that one passes on master as well. Reverting only `languages.json` and `processor/constants.go` to master, `go test ./processor/ -run TestCountStatsDart -v` fails the three behaviour tests with `Expected 4 code got 1` / `Expected 1 comment got 5` / `Expected 1 blank got 0`, `Expected 4 code got 6` / `Expected 1 comment got 0` / `Expected 1 blank got 0`, and `Expected complexity 0 got 3`, and passes the guard.

I checked separately that adding `'` does not disturb the forms built out of it: `'''` and `"""` multi line strings, `r'...'` raw strings and an ordinary Flutter widget file count identically before and after.

`go test -p 4 ./...` and `go test -race ./...` are green on Go 1.26.4 darwin/arm64, `gofmt -l` outside `vendor/` prints nothing, `go vet ./...` is clean for the host and for `GOOS=windows` and `GOOS=linux`, and a second `go generate` leaves `processor/constants.go` byte for byte identical.


boyter:
TY!

PR #774, opened 2026-09-10
docs: add limit and cognitive to the MCP analyze parameter table

NotAFlightRisk:
The `hotspots` and `coupling` tables each document their own `limit`, but `analyze` has the same parameter and is the only table that doesn't mention it. It looks like an oversight, and the published v4.1.0 binary, pointed at this repo's own `processor/` at that tag, confirms it matters.

- With `by_file` enabled, Go reports `"files": 91` but returns 10 entries beacuse `limit` defaults to 10 per language.
- Setting `limit` to `-1` returns all 91, while `limit` set to 3 returns 3.
- Enabling `cognitive` gives Go `"cognitive": 13043`; without it, the field is still present and sat at `0`.

This hits anyone driving scc through MCP, which in practise means an LLM client using the table to decide what it can pass. You ask for a per-file breakdown and get ten files beside a count of 91, with no documented cutoff or way to lift it. Thats a fairly convincing way to think you've got the whole codebase when youve only got ten files per language.

The `hotspots` and `coupling` tables stay exactly as they are. Nothing outside the `analyze` block moves, and no behaviour changes - this is only the docs half of the mismatch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1649,10 +1649,12 @@ The MCP server exposes three tools:
 | `path` | string | no | Directory or file path to analyze. Defaults to current directory. |
 | `sort` | string | no | Column to sort by: `files`, `name`, `lines`, `blanks`, `code`, `comments`, `complexity`, `bytes`. Default: `files`. |
 | `by_file` | boolean | no | If true, return per-file results instead of per-language summary. |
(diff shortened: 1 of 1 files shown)


boyter:
Thanks. Ill try this out as I had not hit it myself, but it could be my setup.

PR #775, opened 2026-09-14
feat: support Txtar and correct TOML information

apocelipes:
Txtar is a simple text-based format to store multiple files together. gofumpt and other projects use it as test data. See <https://pkg.go.dev/golang.org/x/tools/txtar#hdr-Txtar_format> for more info.

In addition, the language information for TOML has been corrected: TOML does not have control structures, and single quotes invalidate escape characters.


boyter:
TY!

PR #778, opened 2026-09-21
docs: add coupling to the --report-skip section list

NotAFlightRisk:
I went to drop the coupling section out of `scc --report` and the table in the README says you cant - every other section is in that list, coupling isn't. It works fine though, the table is just short a name. Passing a bogus section prints the real set:

```
$ scc --report=/tmp/r.html --report-skip bogus .
warning: --report-skip: unknown section "bogus" (recognised: cocomo, locomo, hotspots, coupling, authors, timeline, files, uloc, linelength, card)
```

The rest of the README has it right - the `-h` paste further up lists all ten, and the sentence directly under the table names coupling as one of the git-history sections. So it's only the table thats out of step. I've added it after `hotspots`, which is where the help puts it.

---

The same nine names in the same order also turn up in `TestParseReportSkipKnownNames`, under a comment saying it parses every recognised one. `TestReportSkipRecognisedListMatchesSpec` further down does have all ten and locks the size, so the set itself can't drift. Left that alone, it's a test rather than a doc.

---


boyter:
Cheers. So many options im starting to lose track.

NotAFlightRisk:
>  Cheers. So many options im starting to lose track.

I think I've just noticed another too: ` --list-counters` is missing from readme. I can open aonther PR, but I'll check through each option first to make sure there's nothing else. So that I'm not bombarding you with trivial 1-line PR spam. 

For where the options are defined in the code (e.g. report.go, config.go, report_renderer.go), do you think it could be worth collapsing into a shared source of truth (I guess that would be report.go)?

PR #779, opened 2026-09-22
test: check the README help block against scc --help

NotAFlightRisk:
Hey @boyter, feel free to close this if you don't think it's useful. But following on from #778 I thought it might be helpful to add a drift check so that the readme stays in sync with the actual CLI flags from the code. I've also added the missing ones into the readme while I was here.

`TestReadmeHelpMatchesBinary` runs `scc --help` and compares it against the paste under `$ scc -h`.
It skips the Version, blanks and numeric defaults which change dependening on the system.
If something doesn't match, then it fails and lists the mismatch.

Also from doing this I found some which had drifted, and I fixed them while I was here:
- `--list-counters` wasn't in the paste at all
- `--exp-per-language-counters` still said "currently C, C Header and Java". 


boyter:
Going to take it, and if it works out, keep it.

